PR #1, opened 2015-04-01
IDEX-2188 Fixed actions sorting algorythm and added tests

mkuznyetsov:
Created new sorting algorythm for action groups, that will correctly handle "BEFORE" and "AFTER" constraints


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/action/DefaultActionGroup.java
```diff
@@ -30,12 +33,21 @@
...
+    private static final Action[] EMPTY_ACTIONS = new Action[0];
+
+    /** Contains actions */
+    private final List<Action> actionList = new ArrayList<>();
+
+    /** Contains constraints */
+    private final List<Constraints> constraintsList = new ArrayList<>();
+
+    /** Contains instances of sorted Actions */
+    private Action[] mySortedActions;
```
vzhukovskii: don't call variables with my* our*. define normal name

PR #6, opened 2015-04-08
IDEX-2151. New implementation of sort Parts on View.  Fixed: Size of parts is incorrect after removing the first part from stack.  Added test for PartStackPresenter.

RomanNikitenko:
New implementation of sort Parts on View. 
Fixed: Size of parts is incorrect after removing the first part from stack. 
Added test for PartStackPresenter.


vparfonov:
merge manually  https://github.com/codenvy/che-core/commit/0909a4675795b64b48b69f42d8935a4b5393f8a0

PR #19, opened 2015-04-15
add strategy that makes possible to setup any particular local directory...

vparfonov:
... as root folder for vfs


vparfonov:
megre manually

PR #21, opened 2015-04-16
Cldide 2331

mshaposhnik:


platform-api-client-gwt/che-core-client-gwt-factory/src/main/java/org/eclipse/che/api/factory/gwt/client/FactoryServiceClient.java
```diff
@@ -26,14 +26,11 @@
      *
      * @param raw
      *         factory ID or query string which represents factory non encoded version
```
sleshchenko: please fix java doc

sleshchenko:
Other OK

skabashnyuk:
ok for me

PR #28, opened 2015-04-20
Moved methods which are needed for hosted version directly into implementations

sleshchenko:
@skabashnyuk


skabashnyuk:
ok

sleshchenko:
merged into master with --squash option https://github.com/codenvy/che-core/commit/f05573d03a43abd1ea131c5042004e1aa0956bc0

PR #41, opened 2015-04-30
IDEX-2499 getRunnerList return only runner that respond to the state req...

skabashnyuk:
getRunnerList used in many places of RunQueue. We should not return runners what doesnt respond.


vparfonov:
ok

PR #54, opened 2015-05-07
IDEX-2542 Allow to check user roles for system scope

benoitf:


skabashnyuk:
Is this method in future going to check something else then roles "user", "temp_user", system/*?

benoitf:
It should check account and workspace scope as well. Only problem I see is if I can reuse dao of workspace here for example

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -312,6 +314,57 @@ public void remove(@ApiParam(value = "User ID")
...
+     *         when unable to perform the check
+     */
+    @ApiOperation(value = "Check role for the authenticated user",
+            notes = "Check if user has a role in given scope (default is system) and with an optional scope id. Roles allowed: user, system/admin, system/manager.",
+            response = UserInRoleDescriptor.class,
+            position = 7)
+    @ApiResponses({@ApiResponse(code = 200, message = "OK"),
+                   @ApiResponse(code = 500, message = "Internal Server Error")})
+    @GET
+    @Path("/isUserInRole")
```
garagatyi: Can you change path to something more RESTfull compliant? We are not follow RESTfull method's path naming everywhere, but we want to. It supposes to keep nouns in path. Whereas method type(POST, GET, etc) is a verb.

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -312,6 +314,57 @@ public void remove(@ApiParam(value = "User ID")
...
+    @GenerateLink(rel = LINK_REL_GET_USER_BY_ID)
+    @RolesAllowed({"user", "system/admin", "system/manager"})
+    @Produces(APPLICATION_JSON)
+    public UserInRoleDescriptor isUserInRole(@Required @Description("role inside a scope") @QueryParam("role") String role,
+                                @Description("scope of the role (like system, workspace)") @QueryParam("scope") String scope,
+                                @Description("id used by the scope, like workspaceId for workspace scope") @QueryParam("scopeId") String scopeId,
+                                  @Context SecurityContext context) throws NotFoundException, ServerException {
+        // init values
+        if (scopeId == null){
```
garagatyi: As I remember you can use default value in annotation
`@Default("")`

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -312,6 +314,57 @@ public void remove(@ApiParam(value = "User ID")
...
+            scope = "system";
+        }
+
+        // handle scope
+        boolean isInRole = false;
+        if ("system".equals(scope)) {
+            // check role
+            isInRole = context.isUserInRole(scope + "/" + role);
+        } else {
+            throw new ServerException(String.format("Only system scope is handled for now. Provided scope is %s", scope));
```
garagatyi: I believe we should return forbidden instead of internal server error here. It's not runtime error, can be predicted.

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
benoitf: sure, I was following issue description until now. As we're in user api we can also remove user

so inRole would be ok ? only may let think that we can get all the roles and not check them

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
benoitf: very good remark

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
benoitf: ok, as you prefer

benoitf:
PR updated with comments

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -312,6 +316,52 @@ public void remove(@ApiParam(value = "User ID")
...
+    public UserInRoleDescriptor inRole(@Required @Description("role inside a scope") @QueryParam("role") String role,
+                                       @DefaultValue("system") @Description("scope of the role (like system, workspace)") @QueryParam("scope") String scope,
+                                       @DefaultValue("") @Description("id used by the scope, like workspaceId for workspace scope") @QueryParam("scopeId") String scopeId,
+                                       @Context SecurityContext context) throws NotFoundException, ServerException, ForbiddenException {
+
+        // handle scope
+        boolean isInRole = false;
+        if ("system".equals(scope)) {
+            // check role
+            isInRole = context.isUserInRole(scope + "/" + role);
```
garagatyi: we know scope here because of if statement, so no need to concat scope variable with slash

benoitf: ? we concatenate scope as isUserInRole method is checking for "system/admin", not only ("admin")

benoitf: or you mean you want "system/" + role ?

garagatyi: yes

benoitf:
PR updated

skabashnyuk:
Proposal
1. Check user and temp_user roles
2. Add @Beta annotation for method, at least for couple releases.
3. Implement account and workspace scope, or at least declare that supported only default, system, account and workspace.

wdyt?

benoitf:
#1 how user/temp_user can be checked it's in which scope ?
#2 ok
#3 I asked on the slack chat. Problem for checking account and workspace scope is how I can use these dao from user API ? I can add a dependency from maven mode user to maven account and workspace modules ?

benoitf:
so for 1, user is mapped to only 'user' without scope
based on Gennady specification, default scope was system

It would result in "" being the default and checking admin will use the system scope

so for checking user it will use ("", 'user') while for admin it will use ('system', 'admin')

ok ?

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -312,6 +316,52 @@ public void remove(@ApiParam(value = "User ID")
...
+     */
+    @ApiOperation(value = "Check role for the authenticated user",
+            notes = "Check if user has a role in given scope (default is system) and with an optional scope id. Roles allowed: user, system/admin, system/manager.",
+            response = UserInRoleDescriptor.class,
+            position = 7)
+    @ApiResponses({@ApiResponse(code = 200, message = "OK"),
+                   @ApiResponse(code = 403, message = "Unable to check for the given scope"),
+                   @ApiResponse(code = 500, message = "Internal Server Error")})
+    @GET
+    @Path("/inRole")
```
evoevodin: Use lowercase path instead

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -312,6 +316,52 @@ public void remove(@ApiParam(value = "User ID")
...
+    @GenerateLink(rel = LINK_REL_GET_USER_BY_ID)
```
evoevodin: link rel should be different

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -312,6 +316,52 @@ public void remove(@ApiParam(value = "User ID")
...
+    @RolesAllowed({"user", "system/admin", "system/manager"})
+    @Produces(APPLICATION_JSON)
+    public UserInRoleDescriptor inRole(@Required @Description("role inside a scope") @QueryParam("role") String role,
+                                       @DefaultValue("system") @Description("scope of the role (like system, workspace)") @QueryParam("scope") String scope,
+                                       @DefaultValue("") @Description("id used by the scope, like workspaceId for workspace scope") @QueryParam("scopeId") String scopeId,
+                                       @Context SecurityContext context) throws NotFoundException, ServerException, ForbiddenException {
+
+        // handle scope
+        boolean isInRole = false;
+        if ("system".equals(scope)) {
```
evoevodin: You can simplify it using negative condition

```java

if (!"system".equals(scope)) {
    throw new ForbiddenException(String.format("Only system scope is handled for now. Provided scope is %s", scope));
}

return DtoFactory.getInstance()
                 .createDto(UserInRoleDescriptor.class)
                 .withIsInRole(context.isUserInRole("system/" + role))
                 .withRoleName(role)
                 .withScope(scope)
                 .withScopeId(scopeId);
```

benoitf: yes but idea was to implement all scopes as well

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/shared/dto/UserInRoleDescriptor.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Defines if the current user is or not in a current role
+ * @author Florent Benoit
+ */
+@DTO
+public interface UserInRoleDescriptor {
+
+    @ApiModelProperty(value = "Is in Role")
+    boolean getIsInRole();
```
evoevodin: you should use <code>isInRole()</code> instead

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
benoitf: ok

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
benoitf: ok

platform-api/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/UserServiceTest.java
```diff
@@ -269,6 +271,65 @@ public void shouldBeAbleToRemoveUser() throws Exception {
         verify(userDao).remove(testUser.getId());
     }
 
+
+    /**
+     * Check admin user is 'true' for isUserInRole' with admin role
+     * @throws Exception
+     */
```
evoevodin: Do we really need javadoc here, i think that method declaration describes it enough

skabashnyuk:
how user/temp_user can be checked it's in which scope ?
Default or system. But you asking context role without scope. I know its strange but it is as it is.
About 3. we thinking

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/shared/dto/UserInRoleDescriptor.java
benoitf: https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/shared/dto/AttributeDescriptor.java#L36-42

DTO engine is not handling is pattern from what I saw

platform-api/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/UserServiceTest.java
benoitf: I never see someone complaining about javadoc ? Do you have checked sonar result ?

benoitf:
so I should remove "system" as being the default scope.

default scope will be the empty scope. Then we can check "user" with "" scope

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -312,6 +316,52 @@ public void remove(@ApiParam(value = "User ID")
...
+    @ApiOperation(value = "Check role for the authenticated user",
+            notes = "Check if user has a role in given scope (default is system) and with an optional scope id. Roles allowed: user, system/admin, system/manager.",
+            response = UserInRoleDescriptor.class,
+            position = 7)
+    @ApiResponses({@ApiResponse(code = 200, message = "OK"),
+                   @ApiResponse(code = 403, message = "Unable to check for the given scope"),
+                   @ApiResponse(code = 500, message = "Internal Server Error")})
+    @GET
+    @Path("/inrole")
+    @GenerateLink(rel = LINK_REL_GET_USER_BY_ID)
```
skabashnyuk: Wrong link

skabashnyuk:
>default scope will be the empty scope. Then we can check "user" with "" scope
No. I've talked with @gazarenkov  he thinks that we should consider
  user and temp_user as system scope roles and use it as default.

benoitf:
ok so SSoUser that is being checked has roles which are 'user' or 'system/admin'

In the case we give system scope with user, I will have to check on securityContext.isUserInRole('user') and not securityContext.isUserInRole('system/user')  then

Or you're planning to change 'user' to be 'system/user' ?

benoitf:
Updated PR so user and temp_user can be checked with system scope

benoitf:
to sum up new additional stuff:
 - user and temp_user can be checked with "system" scope (or default one)
 - added @Beta
about checking other scope like workspace or account I think it can be added later as we don't change the pattern.

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/shared/dto/UserInRoleDescriptor.java
evoevodin: DTO engine [handles it](https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/shared/dto/WorkspaceDescriptor.java#L41-L46) as well

benoitf:
Now as we reached more than 30 comments on this PR I think best way should to directly edit my branch to spare exchanges

PR #56, opened 2015-05-11
IDEX-2536: Added Show Hidden Files Persistence Component for saving i…

AndrienkoAleksandr:
Added Show Hidden Files Persistence Component for saving information about hidden files. If user activates show hidden files this component will remember user's choice


vparfonov:
ok

PR #77, opened 2015-05-24
Fixes that allow to search 1. for special characters 2. contains query

rimasirich:
Hello,

I would like to enhance Lucene searcher capabilities therefore I did following changes:
1. queryParser.setAllowLeadingWildcard(true) enables performing contains query ( like *uer* )
2. SimpleAnalyzer uses LetterTokenizer with LowerCaseFilter, that means that it creates tokens by broking a text by non-letter characters, therefore in current implementation it is not possible to search for the text containing special characters. Using WhitespaceTokenizer solves this issue as it creates tokens by broking a text by whitespace.

Thanks & Regards,
Rima


skabashnyuk:
For 1 idea is to allow you to set up custom QueryParser by extending Lucene earcher .  QueryParser will be constucted new for  each search request.
In that custom QueryParser you can set setAllowLeadingWildcard  depending on you busines logic. Will that work for you?

PR #83, opened 2015-05-27
Added workspace id to user preference json

AndrienkoAleksandr:


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/statepersistance/dto/RecentlyProject.java
```diff
@@ -0,0 +1,28 @@
...
+package org.eclipse.che.ide.statepersistance.dto;
+
+import org.eclipse.che.dto.shared.DTO;
+
+/**
+ *
+ * @author Alexander Andrienko
+ */
+@DTO
+public interface RecentlyProject {
```
skabashnyuk: Class name is not good.

PR #89, opened 2015-05-30
Rename the project folder from ".codenvy" to ".che"

liorokman:
Since Che is an Eclipse project, the project folder for projects and modules created with it should not be ".codenvy", but rather ".che"


skabashnyuk:
Unfortunately we are not ready to accept it for 3.x. But we will use it as a base for 4.x. Thanks for contribution.

PR #90, opened 2015-06-01
add support for include files in getTree() of project-api

i072623:
Hello,

Currently, the getTree() api does not include file children in it's response (only folders); to get the files we will need to traverse the tree by calling getChildren() api repeatedly.

So, in order for us to have the functionality we need, I would like to suggest a small addition to the getTree() api (GET /project/{ws-id}/tree/{parent:.*})

I would like to add another query param: ?includeFiles (which will be set to false by default); this param will indicate whether or not to include file children in the result tree.

I've made the relevant changes in: 
platform-api\che-core-api-project\src\main\java\org\eclipse\che\api\project\server\ProjectService.java
and in:
platform-api\che-core-api-project\src\main\java\org\eclipse\che\api\project\server\FolderEntry.java

I've tested it locally; it seems to work fine. I have also added relevant tests to: 
platform-api\che-core-api-project\src\test\java\org\eclipse\che\api\project\server\ProjectServiceTest.java  

Thanks in advance,
Anat

### DIFF
--- a/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/FolderEntry.java
+++ b/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/FolderEntry.java
@@ -44,6 +44,13 @@ public class FolderEntry extends VirtualFileEntry {
         }
     };
 
+    private static final VirtualFileFilter FILE_FOLDER_FILTER = new VirtualFileFilter() {
+        @Override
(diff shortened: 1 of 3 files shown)


vparfonov:
ok for me

skabashnyuk:
ok for me

benoitf:
there are tabs characters and some spaces are missing around parenthesis

PR #94, opened 2015-06-02
Added tests for Lucene Searcher

rimasirich:
Hi,

Following my previous pull request https://github.com/codenvy/che-core/pull/77, I added tests for Lucene Searcher that test: 
1. contains query 
2. query as a regular expression 
3. query that contains special characters

Thanks,
Rima


benoitf:
it could be nice if PR could be rebased before submitting it (so commits has only commit 3df9923)

skabashnyuk:
+1

PR #95, opened 2015-06-02
IDEX-2578: Developed machine perspective view.

dimasnurenko:
@vparfonov @azatsarynnyy


platform-api-client-gwt/che-core-client-gwt-machine/src/main/java/org/eclipse/che/api/machine/gwt/client/MachineServiceClient.java
```diff
@@ -73,6 +74,14 @@
     Promise<List<MachineDescriptor>> getMachines(@Nullable String projectPath);
 
     /**
+     * Find processes related to the specified machine.
+     *
+     * @param machineId
+     *         machine id which allows defines needed machine
```
azatsarynnyy: strange description )
I guess something like: "id of machine to get processes information from" will be better

PR #96, opened 2015-06-02
IDEX-2613; add rest sevices list;

mshaposhnik:


skabashnyuk:
ok. Merge with squash

PR #113, opened 2015-06-11
Copy move with overwrite

Ori-Libhaber:
Hello,

I created this pull request to address the issue described in my previous message to the mailing list
regarding the need to enable copy/move with rename operations to be conducted by a single REST call.

Summary of changes:

* Changes were made to both Copy and Move service methods in che-core-api-project: an additional argument was added, CopyOptions/MoveOptions respectively.
* The body of both services is now in the form of (application/json) { “name” : [the new name to set] , “overwrite” : [true/false] }
* The inclusion of this body is not mandatory so the API is not broken, it’s absence results in previous system behavior for copy/move services.
* Changes were made to org.eclipse.che.vfs.imple.fs project: To be able to pass along the new name and overwrite options, key methods and interfaces were both added and altered, keeping previous behavior intact.
* Tests were added accordingly

Please review and comment.

B.R
Ori Libhaber

### DIFF
--- a/che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
+++ b/che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
@@ -64,6 +64,7 @@ import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
+import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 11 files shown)


skabashnyuk:
It's very hard to see where actual changes and where is changing in source code formatting.
1. Could you plz rollback code formatting what you've added
2. Remove changes in .gitignore
3. Update your PR to latest source code because it's out of date and can't be merged.

Ori-Libhaber:
Hello Sergii,

I pushed another commit which addresses your comments.
Please review and commit.

Best regards,
Ori Libhaber

PR #116, opened 2015-06-14
Support DTOs with properties typed 'any'

tareqhs:
1- Serialzie and de-serialize JsonElement properties as raw JSONs
2- Deep-copy them in the generated copy constructors
3- Avoid redundant copying in fromJsonString() and toJson() cases

### DIFF
--- a/platform-api/che-core-api-dto/pom.xml
+++ b/platform-api/che-core-api-dto/pom.xml
@@ -52,6 +52,18 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>com.google.inject</groupId>
+            <artifactId>guice</artifactId>
+            <version>${com.google.code.guice.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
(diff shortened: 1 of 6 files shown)


vparfonov:
can you provide the same for functionality for client side DTO templates (DtoImplClientTemplat.java)

tareqhs:
Hi @vparfonov,
Sorry for delay.
I noticed that the tests do not generate client implementations for the DTOs. I looked into that and noticed that the client generates GWT JSON code, which means JsonElement is not suitable as a cross-environment type. I suggest using Object instead, which evaluate to JSONValue in client and JsonElement in server. Sounds fine?

vparfonov:
Hi @tareqhs looks good i just have two question:
1. Did you make some test with client-side DTO generation?
2. What means "Int" in fromJsonObjectInt(boolean)?

tareqhs:
Hi @vparfonov,
Thanks for the review.
1. The current code has no tests for client code at all, but i did add the client-side generation with relevant dependencies to make everything compiles (which wasn't done before). I would be glad if we lay out tests for all client code.
2. Int just stands for 'Internal', added this to avoid generated overloaded functions because this is problematic in client code (e.g. JS).

PR #119, opened 2015-06-16
IDEX-2636: Added ability define behaviour for main menu and toolbar depending on Perspective

dimasnurenko:
@vparfonov @evidolob @azatsarynnyy


vparfonov:
ok,except not good name 'updatePerspective'

PR #124, opened 2015-06-19
IDEX-2560: Synchronized process rename item, fixed problem with updat…

AndrienkoAleksandr:
IDEX-2560: Synchronized process rename item, fixed problem with update reference package node


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/event/UpdateTreeNodeChildrenEvent.java
```diff
@@ -0,0 +1,43 @@
...
+ * Update children of some parent TreeNode
+ *
+ * @author Alexander Andrienko
+ */
+public class UpdateTreeNodeChildrenEvent extends GwtEvent<UpdateTreeNodeChildrenEventHandler> {
+
+    public static final Type<UpdateTreeNodeChildrenEventHandler> TYPE = new Type<>();
+
+    private final TreeNode<?>         treeNode;
+    private final AsyncCallback<Void> asyncCallback;
```
vzhukovskii: I think it's not a good idea to use callbacks in event. @vparfonov, @evidolob wdyt?

PR #127, opened 2015-06-22
IDEX-2560: Synchronized process rename item, fixed problem with updat…

AndrienkoAleksandr:
Synchronized process rename item, fixed problem with update reference package node, and uncashed opened files after refresh tree


AndrienkoAleksandr:
Need optimization code

PR #139, opened 2015-07-01
IDEX-2627: added abilities: to override interface methods with DTOs, …

evoevodin:
…to extends Hierarchy of interfaces


skabashnyuk:
norm

PR #143, opened 2015-07-05
Parametrized git server URI prefix

lextiz:


lextiz:
Rebased to https://github.com/codenvy/che-core/pull/149

PR #166, opened 2015-07-22
Ssh script provider

akorneta:


che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/ssh/GitSshScript.java
```diff
@@ -0,0 +1,122 @@
...
+    private byte[] sshKey;
+    private String host;
+    private File   rootFolder;
+    private File   sshScriptFile;
+
+    public GitSshScript(String host, byte[] sshKey) throws GitException {
+        this.rootFolder = Files.createTempDir();
+        this.host = host;
+        this.sshKey = sshKey;
+        sshScriptFile = storeSshScript(writePrivateKeyFile().getPath());
```
skabashnyuk: this.sshScriptFile = storeSshScript(writePrivateKeyFile().getPath());

PR #168, opened 2015-07-24
IDEX-2732: Fixed incorrected message when we delete module from maven…

AndrienkoAleksandr:
Fixed incorrected message when we delete module from maven project


azatsarynnyy:
ok

PR #170, opened 2015-07-24
IDEX-2763 Fixed updating of user's preferences

sleshchenko:


vparfonov:
ok
same for 4.0 branch

PR #177, opened 2015-08-03
IDEX-2619:  Added ability to refresh project tree after changing on file system

dimasnurenko:
@evidolob @vparfonov @skabashnyuk


platform-api/che-core-api-vfs/src/main/java/org/eclipse/che/api/vfs/server/VirtualFile.java
```diff
@@ -592,4 +607,7 @@ VirtualFile createFile(String name, String mediaType, InputStream content)
      *         if any error occurs
      */
     LazyIterator<Pair<String, String>> countMd5Sums() throws ServerException;
+
+    /** Returns instance of {@link java.io.File} */
+    File getIoFile();
```
skabashnyuk: -1 for this method

PR #199, opened 2015-08-21
IDEX-2886: added method to read specific file content to machine API

akorneta:


platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -803,6 +790,36 @@ private LineConsumer getLogger(LineConsumer fileLogger, String outputChannel) th
         return fileLogger;
     }
 
+    private void exec(final InstanceProcess instanceProcess, final String machineId, final LineConsumer processLogger) {
```
evoevodin: document this

skabashnyuk: +

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -331,6 +334,24 @@ public void getProcessLogs(@PathParam("machineId") String machineId,
...
+    @Path("/{machineId}/filepath/{path:.*}")
+    @Produces(MediaType.TEXT_PLAIN)
+    @RolesAllowed("user")
+    public String getFileContent(@PathParam("machineId") String machineId,
+                                 @PathParam("path") String path)
+            throws NotFoundException, ForbiddenException, ServerException, IOException {
+        checkCurrentUserPermissions(machineManager.getMachine(machineId));
+
+        CommandDescriptor descriptor = newDto(CommandDescriptor.class)
+                .withCommandLine(String.format("cat %1$2s", path));
```
evoevodin: i think you can define it in one line and without of external formatting 
<code> CommandDescriptor descriptor = newDto(CommandDescriptor.class).withCommandLine("cat " + path)</code>

garagatyi: +1

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -582,42 +582,28 @@ public void removeSnapshots(String owner, String workspaceId, ProjectBinding pro
      */
     public InstanceProcess exec(final String machineId, final Command command, String clientOutputChannel)
             throws NotFoundException, MachineException {
+        return exec(machineId, command, clientOutputChannel, null, true);
+    }
+
+    public InstanceProcess exec(final String machineId,
+                                final Command command,
+                                final String clientOutputChannel,
```
garagatyi: why this method contains both processLogger and clientOutputChannel?
I propose separate it into 2 methods. First accepts only process logger. Another one accepts only client output channel, creates process logger and calls the first one.
Wdyt?

Issue #206, opened 2015-08-25
Posibility for optimisation in ParameterizedTypeImpl.java 


alexVengrovsk:
After looking to ```ParameterizedTypeImpl.java```, it is not very clear, why there are so many private constructors? It is posible to use only two constructors
```
private ParameterizedTypeImpl(Class<?> rawType, Type typeArgument) {
        this.rawType = rawType;
        this.typeArguments = new Type[]{typeArgument};
    }

private ParameterizedTypeImpl(Class<?> rawType, Type[] typeArguments) {
        this.rawType = rawType;
        this.typeArguments = new Type[typeArguments.length];
        System.arraycopy(typeArguments, 0, this.typeArguments, 0, this.typeArguments.length);
    }
```
as any of stacic generation methods (```newParameterizedType(Class<?> rawType, Type... typeArguments) and newParameterizedType(Class<?> rawType, Type typeArgument)```) doesn't provide information for ```ownerType``` variable instantiation. According to current version of the class, this variable is allways ```null```. But it will be ```null``` by default.

All these three-parameter constructors is useful only, when one more method is added.
For example:
```
public static ParameterizedType newParameterizedType(Type ownerType, 
   Class<?> rawType, Type typeArgument) {
   return new ParameterizedTypeImpl(ownerType, rawType, typeArgument);
}
```

skabashnyuk:
Are you ready to provide PR?

alexVengrovsk:
Yes. I can do it today

alexVengrovsk:
I added a PR. [#208](https://github.com/codenvy/che-core/pull/208)

PR #208, opened 2015-08-26
[#206 - issue] Private constructors changed in ParameterizedTypeImpl.java

alexVengrovsk:
According to issue #206, we can remove code overhead by making these changes.

But, it is still a question - Will this class provide the ```ownerType``` variable's instantiation through static generation methods. If so, than all these changes are inappropriate, and the different approach should be used (there should be static generation methods with three parameters):
```
public static ParameterizedType newParameterizedType(Type ownerType, 
   Class<?> rawType, Type typeArgument) {
   return new ParameterizedTypeImpl(ownerType, rawType, typeArgument);
} 
```
and
```
public static ParameterizedType newParameterizedType(Type ownerType, 
   Class<?> rawType, Type... typeArgument) {
   return new ParameterizedTypeImpl(ownerType, rawType, typeArgument);
} 
```
And class will need its ```private final Type ownerType``` variable back, and two constructors(everyone with three parameters)

### DIFF
--- a/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/reflect/ParameterizedTypeImpl.java
+++ b/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/reflect/ParameterizedTypeImpl.java
@@ -21,14 +21,10 @@ import java.util.Arrays;
  */
 public final class ParameterizedTypeImpl implements ParameterizedType {
 
-    public static ParameterizedType newParameterizedType(Class<?> rawType, Type... typeArguments) {
-        return new ParameterizedTypeImpl(rawType, typeArguments);
-    }
-
     private final Type[]   typeArguments;
     private final Class<?> rawType;
 
-    private ParameterizedTypeImpl(Class<?> rawType, Type[] typeArguments) {
+    public ParameterizedTypeImpl(Class<?> rawType, Type... typeArguments) {
         this.rawType = rawType;
         this.typeArguments = new Type[typeArguments.length];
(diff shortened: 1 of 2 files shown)


benoitf:
I would say that one constructor should still call the other one

benoitf:
with varags, there could have only one constructor
ParameterizedTypeImpl(Class<?> rawType, Type... types)

as well

    public static ParameterizedType newParameterizedType(Class<?> rawType, Type... typeArguments)
is redundant with
    public static ParameterizedType newParameterizedType(Class<?> rawType, Type typeArgument)

as for the method we can give only one argument if we want

alexVengrovsk:
You are right. I can change it to one constructor only.

benoitf:
Now I'm thinking that we may also think why we need to use a static method while we could use directly the constructor

alexVengrovsk:
Yes, it is a question to think about. I dont see any help from static methods. We don't create a singleton here.
Instances of this class were created only twice in DtoFactory. java. We can easily replace static methods and use just constructors.

benoitf:
@skabashnyuk WDYT ?

alexVengrovsk:
Please, check my new commit

benoitf:
looks like a lot simpler for me

skabashnyuk:
Alex did you test che  assembly with this changes?

alexVengrovsk:
No. I didn't perform any additional testing

benoitf:
@skabashnyuk I've rebuild che assembly with Alex changes and it's OK on my side

alexVengrovsk:
@benoitf, thank you for testing my changes!

PR #210, opened 2015-08-26
Bug fix: populate workspace and project fields of BuildTaskDescriptor

tareqhs:
Change-Id: Icf9cf5db9b6cdc768a55c7fd31e123458833a6a7


skabashnyuk:
Could you plz inline methid withCommonDescriptorProperties

tareqhs:
hi @skabashnyuk,
wouldn't inlining it make the operation less maintainable?

skabashnyuk:
Author of BuildQueueTask and I think that it's better to have it a bit less maintainable, than one small private method ))

benoitf:
I think Tareq proposal is more maintainable as well. If new methods are added, all attributes will be correctly setup by calling this single method

PR #213, opened 2015-08-27
IDEX-2886: added method to read specific file content to machine API

akorneta:


platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -72,7 +76,7 @@
     @Inject
     public MachineService(MachineManager machineManager) {
         this.machineManager = machineManager;
-        this.dtoFactory = DtoFactory.getInstance();
+        this.dtoFactory = getInstance();
```
garagatyi: remove this string

garagatyi:
PR contains too much changes that are not related to the topic. It's hard to analyze PR because of that.
@akorneta please next time make such changes in source branch and merge to your branch to help reviewers reduce time needed to make review. I know that we often do the same, but we all should try to avoid that.

skabashnyuk:
Javadoc plz

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -701,6 +701,81 @@ public void run() {
...
+
+    /**
+     * Gets machines states.
+     *
+     * @return list of machine states
+     * @throws MachineException
+     *         if any error occur with machine registry
+     */
+    public List<MachineImpl> getMachinesStates() throws MachineException {
+        return machineRegistry.getStates();
```
skabashnyuk: Is this a copy or reference?

akorneta: references

PR #216, opened 2015-08-28
IDEX-2180:added an ability to save git committer name and email into user prefe rences

svor:


svor:
Merged into 4.0. [Git commit](https://github.com/codenvy/che-core/commit/09b3cbb1ac41af20a99e55c141f068ac752d5a15)

PR #218, opened 2015-08-30
Update Selection.java (and code, that use it's deprecated method), SetupPomAndGWTModule.java

alexVengrovsk:
Delete deprecated method ```getFirstElement()``` from ```Selection.java```.
Change it for ```getHeadElement()``` in code.


skabashnyuk:
@vparfonov  reported to me  that he is not ok with  https://github.com/codenvy/che-core/pull/218/files#diff-d00b260c79d6bd602c9d32049edbb1feR18

Also you need to update this pr because of merge confilcts

PR #219, opened 2015-08-31
Update CronExpression.java

alexVengrovsk:
Change ```ParseException``` in catch clause of method ```isValidExpression(String cronExpression)``` to ```Exception``` to handle ```IllegalArgumentException``` too.


skabashnyuk:
I'm not willing to change without the strong reason original quartz code.

alexVengrovsk:
@skabashnyuk It is up to you. I understand this.

PR #220, opened 2015-08-31
Bug fix: synchronize preferences update

tal-sapan:
Since UserProfileService updates the user preferences by merging the sent changes with the existing preferences, it should be synchronized during those actions so that 2 calls for modifying preferences executed at the same time don't cause data inconsistency.

Change-Id: Ifc3b95c07a3a0d007b91e5e9efae2627fb3f169f


tal-sapan:
It looks like the build failed due to reasons not related to my change - I will try to retrigger it by closing and re-opening the pull request.

tal-sapan:
Sorry, looks like I mixed in an unrelated commit with my change, trying to figure out how to fix it...

tal-sapan:
I created pull request https://github.com/codenvy/che-core/pull/223 with the updated change instead of this one

PR #221, opened 2015-08-31
IDEX-2899. Added RemoveModuleHandler for delete module from pom.xml.Refresh content of '.codenvy/modules' and 'pom.xml'.

RomanNikitenko:
Added RemoveModuleHandler for delete module from pom.xml.
Refresh content of '.codenvy/modules' and 'pom.xml' at removing module.


RomanNikitenko:
merged manually

PR #223, opened 2015-08-31
Bug fix: synchronize preferences update

tal-sapan:
See https://github.com/codenvy/che-core/pull/220

### DIFF
--- a/platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserProfileService.java
+++ b/platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserProfileService.java
@@ -29,6 +29,7 @@ import org.eclipse.che.api.user.shared.dto.ProfileDescriptor;
 import org.eclipse.che.commons.env.EnvironmentContext;
 import org.eclipse.che.dto.server.DtoFactory;
 
+import com.google.common.util.concurrent.Striped;
 import com.wordnik.swagger.annotations.Api;
 import com.wordnik.swagger.annotations.ApiOperation;
 import com.wordnik.swagger.annotations.ApiParam;
@@ -69,6 +70,8 @@ import static org.eclipse.che.api.user.server.Constants.LINK_REL_UPDATE_USER_PRO
 import static com.google.common.base.Strings.nullToEmpty;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
+import java.util.concurrent.locks.Lock;
+
 /**
  * User Profile API
  *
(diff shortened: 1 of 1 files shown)


skabashnyuk:
I assume @evoevodin   proposed to use this lock http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html

tal-sapan:
@skabashnyuk The lock you linked to is an interface. I'm not aware of any implementations that take a parameter (except what I linked on an answer in stackoverflow: http://stackoverflow.com/questions/12450402/java-synchronizing-based-on-a-parameter, which is not tested).
In this question and in other discussions the solutions are pretty much divided to 2: either use String.intern() or implement a parameter-based lock (using a string-to-lock map). 

The intern solution has its drawbacks, mainly that another part of the code might accidently lock on the same string which could cause a deadlock (the other drawback that it could use a lot of memory is not relevant since these strings are garbage collected just like in a WeakHashMap, especially in Java 7 and above since they are kept on the heap). For this reason I added a prefix to the user name. I don't think there's a risk of deadlock now with this change (it's possible to also add the class name to make it more unique).

The parameter-based lock implementation is much more complex since you need to make sure the operations on the map are thread safe and prevent race conditions that might occur. It's more bug-prone and difficult to test. I think it adds unnecessary complexity to the solution.

By the way, why do I get this broken pipe error message every other build? It looks like a sporadic failure...

skabashnyuk:
What about this ?
http://codingjunkie.net/striped-concurrency/ 
https://dzone.com/articles/fine-grained-concurrency-guava
http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/util/concurrent/Striped.html

tal-sapan:
That's interesting, although it still requires to set a maximum number of available locks, which could differ depends on the resources available to the Che implementation. I suppose it can be set to a big number (10000?) and use the lazy version but I don't know what this would impact. What do you think?

skabashnyuk:
10000 I think it's too optimistic.  I would say 100 or 1000 at max. 
Weak and lazy - wdyt?

tal-sapan:
According to the documentation there are 2 options to create the locks - either by strong reference, in which case they are all created eagerly (when the Striped object is initialized) and are kept in memory until the Striped object is released (i.e. not while the server is running in this case) or by weak reference, in which case they are created on-demand with weak references and are only kept in memory while there is a reference to them outside the Striped object (this prevents too big memory footprint by allowing them to be garbage collected). I think the second option is better suited to this use case.
Depending on the inner implementation of the lazy Striped object, if it doesn't affect the memory footprint until you actually ask for a lock, setting a high number could be preferable (also because it would lower the possibility of 2 users getting the same lock and then the second user will have to wait until the first user finishes to update their preferences).
I assume the expected number of concurrent users is a good place to start but this could change according to Che implementation.

skabashnyuk:
I think the second option is better suited to this use case.
+

skabashnyuk:
will you rework it?

tal-sapan:
Yes, I will modify it - it might take a few days though due to other priorities we have right now.

tal-sapan:
Hi,

I updated the change to use Striped. Please review.
Also, do you know why the build is failing? I don't think it's related to my change.

benoitf:
I've restarted the build process

benoitf:
I've noticed that tab character is used while 4 spaces should be used.

tal-sapan:
Updated to use spaces instead of tabs

evoevodin:
Looks good, well done. :+1:

PR #225, opened 2015-09-01
IDEX-2972: fixed problems with GitService

svor:


vparfonov:
ok

svor:
[Merge commit](https://github.com/codenvy/che-core/commit/dac9ca39b63b52f8d9b61d05708db9826e5ee2cd)

PR #232, opened 2015-09-04
Idex 2790. Add loader with initialize extensions, start workspace and dev machine information.

RomanNikitenko:
Add loader with initialize extensions, start workspace and dev machine information.


slemeur:
Should we not take into account the issue: https://jira.codenvycorp.com/browse/IDEX-3019 before merging?

vparfonov:
i don\t think so we need to start with something, because for now we don't show any events/logs. In this case hard to understand that goes wrong

vparfonov:
ок

PR #236, opened 2015-09-07
Idex 2909

mshaposhnik:


platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
```diff
@@ -236,6 +239,20 @@ public Response invalidate() throws IOException {
                                                                            null)).type(MediaType.TEXT_PLAIN).build();
     }
 
+    @GET
+    @Produces({MediaType.APPLICATION_JSON})
```
evoevodin: braces are not required in this case

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
```diff
@@ -236,6 +239,20 @@ public Response invalidate() throws IOException {
...
+    @RolesAllowed({"user", "temp_user"})
+    @Path("token")
```
evoevodin: I think <code>@Path</code> should be declared after Request type _(GET)_ in this case, as for me it is more logical sequence, see examples [here](https://github.com/codenvy/che-core/blob/4.0/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java) and  [here](https://github.com/codenvy/che-core/blob/4.0/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/command/CommandService.java)

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
```diff
@@ -236,6 +239,20 @@ public Response invalidate() throws IOException {
...
+    public OAuthToken token(@QueryParam("oauth_provider") String oauth_provider, @Context SecurityContext security)
+            throws Exception {
```
evoevodin: Rename <code>oauth_provider</code> to <code>oauthProvider</code> (actually i am working on codestyle guide - it covers it as well). 
What if <code>oauthProvider</code> is <code>null</code>,  it must be documented!
Why do method throws <code>Exception</code>? It is better to use separate exceptions in this case such as <code>ConflictException</code> or <code>NotFoundException</code> and document it.

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
```diff
@@ -236,6 +239,20 @@ public Response invalidate() throws IOException {
...
+        final Principal principal = security.getUserPrincipal();
+        OAuthAuthenticator provider = getAuthenticator(oauth_provider);
```
evoevodin: Why <code>principal</code> is defined as <code>final</code> and provider is not?
Also why do you need local variable here?

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/RemoteOAuthTokenProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.ws.rs.core.UriBuilder;
+import java.io.IOException;
+
+
+/**
+ * Allow get token from OAuth service over http.
```
evoevodin: Where is <code>@author Maxim Shaposhnik</code> ?

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/RemoteOAuthTokenProvider.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public class RemoteOAuthTokenProvider implements OAuthTokenProvider {
+    private static final Logger LOG = LoggerFactory.getLogger(RemoteOAuthTokenProvider.class);
+
+    @Inject
+    @Named("api.endpoint")
+    protected String apiEndpoint;
+
+    @Override
+    public OAuthToken getToken(String oauthProviderName, String userId) throws IOException {
```
evoevodin: Document it or at least use <code>@inheritDoc</code>

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/RemoteOAuthTokenProvider.java
```diff
@@ -0,0 +1,65 @@
...
+        if (!userId.isEmpty()) {
```
evoevodin: Use non negative check it will help you to avoid nesting, also it looks better :

```java
if (userId.isEmpty()) {
    return null;
}
```

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/RemoteOAuthTokenProvider.java
```diff
@@ -0,0 +1,65 @@
...
+                try {
```
evoevodin: Why this code is indented with 8 spaces instead of 4?

evoevodin:
Can you please add tests?

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/RemoteOAuthTokenProvider.java
```diff
@@ -0,0 +1,65 @@
...
+                    UriBuilder ub = UriBuilder.fromUri(apiEndpoint)
+                                              .path("/oauth/token/")
```
evoevodin: Is it possible to use direct REST method reference here?
If it is you should use it! See [example](https://github.com/codenvy/che-core/blob/4.0/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/command/CommandService.java#L190-L195).
The only reason why it may not be used is when you have cycle dependency in your project.

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
```diff
@@ -236,6 +240,27 @@ public Response invalidate() throws IOException {
...
+     * @param oauthProvider OAuth provider name
+     * @param security Security context
+     * @return OAuthToken
+     * @throws IOException
+     */
+    @GET
+    @Path("token")
+    @Produces(MediaType.APPLICATION_JSON)
+    @RolesAllowed({"user", "temp_user"})
+    public OAuthToken token(@Required @QueryParam("oauth_provider") String oauthProvider, @Context SecurityContext security)
```
mshaposhnik: Security context is already defined as class member

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
```diff
@@ -236,6 +240,23 @@ public Response invalidate() throws IOException {
...
+     * @return OAuthToken
+     * @throws IOException
+     */
+    @GET
+    @Path("token")
+    @Produces(MediaType.APPLICATION_JSON)
+    @RolesAllowed({"user", "temp_user"})
+    public OAuthToken token(@Required @QueryParam("oauth_provider") String oauthProvider)
+            throws IOException {
+        OAuthAuthenticator provider = getAuthenticator(oauthProvider);
```
garagatyi: We are not using WebApplicationException anymore. Please replace it with BadRequestException

PR #237, opened 2015-09-08
IDEX-2784: fixed import project from github on Windows OS

AndrienkoAleksandr:


AndrienkoAleksandr:
Closed because was improved in pull reques https://github.com/codenvy/che-core/pull/245

PR #242, opened 2015-09-09
IDEX-2992. Don't unsubscribe from channel before MachineStatusNotifier gets machine status. Fixed NullPointerException.

RomanNikitenko:
Don't unsubscribe from channel before MachineStatusNotifier gets machine status. 
Fixed NullPointerException.


vparfonov:
ok

PR #243, opened 2015-09-09
Use "mixins" instead of "mixinTypes" in ProjectUpdate JSONs

tareqhs:
this makes the field similar to ProjectDescriptor from the client's POV

currently the project service to uses NewProject and UpdateProejct in the service method that create and update projects, respectively. These DTOs contains the list of mix-in project types in "mixinTypes" field. This is different from ProjectDescriptor DTO which is returned by the service methods that return project metadata, including two service method mentioned earlier, which has a field named "mixins". This leads to confusions and workarounds by the API consumers as they pass project configurations back and forth.

only updateProject() and createProject() service APIs REQUESTS are affected. Existing project JSONs, and accordingly service APIs that return them, are not affected. The responses of updateProject() and createProject() themselves are not affected as well.

Change-Id: I151e6f17499c4003ed7e17c7646f5b5f8f3ee090


sleshchenko:
This pull request will be merged after https://jira.codenvycorp.com/browse/IDEX-3294 is done

mkuznyetsov:
Merged to master as 00aedf6b7a917c89fb4b8f73cb70425a8dc2bb85

PR #244, opened 2015-09-09
IDEX-3018: import projects using WS

svor:


skabashnyuk:
ok for me. What about @vparfonov ?

PR #248, opened 2015-09-10
IDEX-2881: added api methods for copy resources between machines

akorneta:


platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,39 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+     *         path to resources in source machine
+     * @param targetMachineId
+     *         id of target machine
+     * @param targetPath
+     *         destination path in target machine
+     * @param noOverwrite
+     *         If "1", "true", or "True" then it will be an error
+     *         if unpacking the given content would cause an existing directory to be replaced
+     *         with a non-directory and vice versa.
+     * @throws MachineException
```
garagatyi: Method also throws NotFoundException and ForbiddenException but you haven't mentioned it in javadoc

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,39 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+     *
+     * @param sourceMachineId
+     *         id of source machine
+     * @param sourcePath
+     *         path to resources in source machine
+     * @param targetMachineId
+     *         id of target machine
+     * @param targetPath
+     *         destination path in target machine
+     * @param noOverwrite
```
garagatyi: It is ugly. I understand that it is copied from docker API but I don't agree to follow their API

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,39 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+     *         If "1", "true", or "True" then it will be an error
```
garagatyi: I suppose we should support `?overwrite=true|false` only

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,39 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+    @POST
+    @Path("/{machineId}/filepath/{sourcePath:.*}")
+    @RolesAllowed("user")
+    public void transferFilesBetweenMachines(@PathParam("machineId") String sourceMachineId,
+                                             @PathParam("sourcePath") String sourcePath,
+                                             @QueryParam("targetMachineId") String targetMachineId,
+                                             @QueryParam("targetPath") String targetPath,
+                                             @DefaultValue("false") @QueryParam("noOverwrite") Boolean noOverwrite)
+            throws NotFoundException, ServerException, ForbiddenException {
+        checkCurrentUserPermissions(machineManager.getMachine(sourceMachineId));
```
garagatyi: Please add check of user permissions for target machine also

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,39 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+
+        machineManager.getMachine(targetMachineId)
+                      .putResource(machineManager.getMachine(sourceMachineId).getResource(sourcePath), targetPath, noOverwrite);
```
garagatyi: I would rename `putResource` to `copy`. WDYT?

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,39 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+    @RolesAllowed("user")
+    public void transferFilesBetweenMachines(@PathParam("machineId") String sourceMachineId,
+                                             @PathParam("sourcePath") String sourcePath,
+                                             @QueryParam("targetMachineId") String targetMachineId,
+                                             @QueryParam("targetPath") String targetPath,
+                                             @DefaultValue("false") @QueryParam("noOverwrite") Boolean noOverwrite)
+            throws NotFoundException, ServerException, ForbiddenException {
+        checkCurrentUserPermissions(machineManager.getMachine(sourceMachineId));
+
+        machineManager.getMachine(targetMachineId)
```
garagatyi: With this code you declare copying via API instance. What if 
`machineManager.getMachine(machineId).copy(targetMachineId, targetPath, overwrite)`
@akorneta  @skabashnyuk WDYT?

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,39 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+     *         If "1", "true", or "True" then it will be an error
+     *         if unpacking the given content would cause an existing directory to be replaced
+     *         with a non-directory and vice versa.
+     * @throws MachineException
+     *      if any error occurs with coping resources between machines
+     */
+    @POST
+    @Path("/{machineId}/filepath/{sourcePath:.*}")
+    @RolesAllowed("user")
+    public void transferFilesBetweenMachines(@PathParam("machineId") String sourceMachineId,
```
garagatyi: Why not use word `copy ` instead of `transfer`?

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
skabashnyuk: ok

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
skabashnyuk: ok

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,37 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+     *         if any error occurs with getting source machine
+     */
+    @POST
+    @Path("/{machineId}/filepath/{sourcePath:.*}")
+    @RolesAllowed("user")
+    public void copyFilesBetweenMachines(@PathParam("machineId") String sourceMachineId,
+                                         @PathParam("sourcePath") String sourcePath,
+                                         @QueryParam("targetMachineId") String targetMachineId,
+                                         @QueryParam("targetPath") String targetPath,
+                                         @DefaultValue("false") @QueryParam("overwrite") Boolean overwrite)
```
skabashnyuk: sync signature with vfs method copy

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -362,6 +363,37 @@ public String getFileContent(@PathParam("machineId") String machineId,
...
+            throws NotFoundException, ServerException, ForbiddenException {
+        checkCurrentUserPermissions(machineManager.getMachine(sourceMachineId));
+
+        machineManager.getMachine(targetMachineId).copy(machineManager.getMachine(sourceMachineId), sourcePath, targetPath, overwrite);
```
skabashnyuk: Do we need here check for null of input parameters?

akorneta: it is not needed because when machine id null we throw exception machine with that id not found and if path null we have exception from docker api  'path' cannot be empty

skabashnyuk:
ok

skabashnyuk:
ok

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/spi/Instance.java
```diff
@@ -134,4 +135,22 @@
...
+     *         source machine
+     * @param sourcePath
+     *         path to file or directory inside specified machine
+     * @param targetPath
+     *         path to destination file or directory inside machine
+     * @param overwrite
+     *         If "false" then it will be an error if unpacking the given content would cause
+     *         an existing directory to be replaced with a non-directory and vice versa.
+     * @throws MachineException
+     *         if any error occurs when files are copying into the machine
```
garagatyi: are being copied

garagatyi:
ok

PR #251, opened 2015-09-11
IDEX-3007: Fixed problem with update opened files after rename folder

AndrienkoAleksandr:


skabashnyuk:
what happened with test?

vparfonov:
see comment here https://github.com/codenvy/che-core/pull/251/files#diff-d00b260c79d6bd602c9d32049edbb1feR196

PR #255, opened 2015-09-14
IDEX-2784: Fixed import project from github on Windows OS

AndrienkoAleksandr:


vparfonov:
Don't need it in 4.0 version because git server work on Machine side under Linux OS

PR #258, opened 2015-09-15
Fix restore state, added helper method for searching node and little bugfixes

vzhukovskii:
@vparfonov in this PR was recovered restore state(expand, opened files, last active file), open file by factory, added helper method for searching nodes by storable path


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/NewProjectExplorerPresenter.java
```diff
@@ -163,19 +170,25 @@ public void onResourceEvent(ResourceNodeDeletedEvent evt) {
...
-                    @Override
-                    public void setData(@NotNull Object data) {
-
-                    }
-                };
-                reloadChildren(event.getNode().getParent(), dataObject);
+                Object newDataObject = event.getNewDataObject();
+                String path = null;
+
+                if (newDataObject instanceof ProjectReference) {
```
vparfonov: Need to simplify if possible or add some comments because hard to understand this block

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/event/ProjectStateRestoredEvent.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public ProjectDescriptor getDescriptor() {
+        return descriptor;
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    @Override
+    public Type<ProjectStateRestoredHandler> getAssociatedType() {
+        return (Type)TYPE;
```
evoevodin: Why do you need this cast? 
Also why do you need to suppress warnings here?
Is it okay that this method may return <code>null</code>, also why this fact is not documented?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/SearchNodeHandler.java
```diff
@@ -0,0 +1,250 @@
...
+import org.eclipse.che.ide.ui.smartTree.event.PostLoadEvent.PostLoadHandler;
+
+import java.util.List;
+
+/**
+ * Search node handler, perform searching specified node in the tree by storable value.
+ * For example if user passes "/project/path/to/file" then this node handler will check
+ * opened root nodes and if it contains project node with path "/project" then it will
+ * search children by path "path/to/file".
+ * @author Vlad Zhukovskiy
```
evoevodin: It seems that it is not formatted, empty line required before the <code>@author</code> line.

ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/smartTree/event/NodeAddedEvent.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public NodeAddedEvent(List<Node> nodes) {
+        this.nodes = Collections.unmodifiableList(nodes);
+    }
+
+    public NodeAddedEvent(Node node) {
+        nodes = Collections.singletonList(node);
+    }
+
+    /** {@inheritDoc} */
```
azatsarynnyy: No need to inherit comment explicitly with {@inheritDoc} tag if you have empty comment there.
In this case Javadoc tool will copy comment from base class automatically.
http://docs.oracle.com/javase/7/docs/technotes/tools/solaris/javadoc.html#inheritingcomments

PR #259, opened 2015-09-16
IDEX-2989, IDEX-3010: Wrong behaviours with render package nodes and …

vzhukovskii:
…fix related with open file action
@vparfonov


skabashnyuk:
ok. what about @vparfonov ?

vparfonov:
ok, what about tests?

vzhukovskii:
they will be in separate commit

vzhukovskii:
merged with squash https://github.com/codenvy/che-core/commit/e35e022aa869927defb6711fae0228a204818d93

PR #260, opened 2015-09-16
IDEX-3056: fixed module docker runner availability

akorneta:


platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -993,32 +1002,61 @@ private void importRunnerEnvironment(ImportProject importProject, FolderEntry ba
...
+
+        if (!(runnerSourceValue.getLocation().startsWith("https") || runnerSourceValue.getLocation().startsWith("http"))) {
+            LOG.warn("ProjectService.importProject :: not valid runner source location available only http or https scheme but we get :" +
+                     runnerSourceValue);
+            return false;
+        }
+        return true;
+    }
+
+    private void checkDefaultProjectFolder(NewProject newProject) {
```
mshaposhnik: I beleive this method should be used for checking module default runners too

akorneta: +

PR #264, opened 2015-09-17
Remove IDE-version property from manifest

mkuznyetsov:


skabashnyuk:
ок

riuvshin:
Are you sure that we don't need it for 4.0 ? AFAIK this is required for IM.

skabashnyuk:
May be we need some property. But we have no ability to set it here because che-core has no idea or relation to cloud-ide version

PR #268, opened 2015-09-18
Idex 2848

akorneta:


platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -76,7 +77,8 @@
  */
 @Singleton
 public class MachineManager {
-    private static final Logger LOG = LoggerFactory.getLogger(MachineManager.class);
+    private static final Logger  LOG                          = LoggerFactory.getLogger(MachineManager.class);
+    private static final Pattern MACHINE_DISPLAY_NAME_PATTERN = Pattern.compile("^/?[a-zA-Z0-9_-]+$");
```
evoevodin: I think you should explain it [example](https://github.com/codenvy/che-core/blob/4.0/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java#L71)

platform-api/che-core-api-machine/src/test/java/org/eclipse/che/api/machine/server/MachineManagerTest.java
```diff
@@ -0,0 +1,142 @@
...
+        EnvironmentContext.setCurrent(envCont);
+    }
+
+    @AfterMethod
+    public void tearDown() throws Exception {
+        EnvironmentContext.reset();
+    }
+
+    @Test(expectedExceptions = MachineException.class)
+    public void shouldBeAnExceptionWhileValidatingDockerDisplayName() throws Exception {
```
evoevodin: i think that <code>shouldThrow</code> is better here

platform-api/che-core-api-machine/src/test/java/org/eclipse/che/api/machine/server/MachineManagerTest.java
```diff
@@ -0,0 +1,142 @@
...
+        final RecipeMachineCreationMetadata machineCreationMetadata = mock(RecipeMachineCreationMetadata.class);
+        final MachineRecipe machineRecipe = mock(MachineRecipe.class);
+        final InstanceProvider instanceProvider = mock(InstanceProvider.class);
+        final String type = "type";
+        final Set<String> recipeTypes = new HashSet<String>() {{
```
evoevodin: Hmm, actually here you create an instance of anonymous class which is not good, you can simplify it:
<code>new HashSet<>(Collections.singleton(type));

platform-api/che-core-api-machine/src/test/java/org/eclipse/che/api/machine/server/MachineManagerTest.java
```diff
@@ -0,0 +1,142 @@
...
+        envCont.setUser(new UserImpl("user", null, null, null, false));
+        EnvironmentContext.setCurrent(envCont);
+    }
+
+    @AfterMethod
+    public void tearDown() throws Exception {
+        EnvironmentContext.reset();
+    }
+
+    @Test(expectedExceptions = MachineException.class)
```
evoevodin: Consider adding expectedMessagePattern

platform-api/che-core-api-machine/src/test/java/org/eclipse/che/api/machine/server/MachineManagerTest.java
akorneta: :+1:

platform-api/che-core-api-machine/src/test/java/org/eclipse/che/api/machine/server/MachineManagerTest.java
akorneta: yeap

garagatyi:
ok

evoevodin:
:+1:

PR #273, opened 2015-09-21
IDEX-2864: adopt new styles on main menu.

ashumilova:


skabashnyuk:
Need to update this pr
+0

PR #275, opened 2015-09-21
Add sort to IDE plug-ins list in prefs dialog box

TylerJewell:
Added a comparator so that the list of extensions is alphabeticized.


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/extension/ExtensionManagerPresenter.java
```diff
@@ -65,6 +67,10 @@ public void go(AcceptsOneWidget container) {
         for (ExtensionDescription ed : extensionRegistry.getExtensionDescriptions().values()) {
             extensions.add(ed);
         }
+
+        // Sort the list of plug-ins alphabetically
+        Collections.sort(extensions, (ed1, ed2) -> ed1.getTitle().compareTo(ed2.getTitle()));
```
vparfonov: GWT don't support Java 8 syntax

PR #279, opened 2015-09-23
Added posibility refactor java classes to client side

AndrienkoAleksandr:


ide/che-core-ide-jseditor/src/main/java/org/eclipse/che/ide/jseditor/client/texteditor/TextEditor.java
```diff
@@ -84,4 +84,6 @@
      * @param message message
      */
     void showMessage(String message);
+
+    boolean isFocused();
```
skabashnyuk: javadoc

PR #283, opened 2015-09-24
Idex 3060

garagatyi:


platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -105,6 +111,7 @@ public MachineManager(SnapshotDao snapshotDao,
         this.defaultMachineMemorySizeMB = defaultMachineMemorySizeMB;
 
         executor = Executors.newCachedThreadPool(new ThreadFactoryBuilder().setNameFormat("MachineManager-%d").setDaemon(true).build());
+        this.machineCleaner = new MachineCleaner();
```
skabashnyuk: why not injected?

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -872,8 +886,32 @@ private LineConsumer getLogger(LineConsumer fileLogger, String outputChannel) th
         return fileLogger;
     }
 
+    // cleanup machine if event about instance failure comes
+    private class MachineCleaner implements EventSubscriber<InstanceStateEvent> {
```
skabashnyuk: may be separate class?

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -872,8 +886,32 @@ private LineConsumer getLogger(LineConsumer fileLogger, String outputChannel) th
...
+                    if (event.getType() == OOM) {
+                        message = message +
+                                  "The processes in this machine need more RAM. This machine started with " + machine.getMemorySize() +
+                                  "MB. Create a new machine configuration that allocates additional RAM or increase " +
+                                  "the workspace RAM limit in the user dashboard.";
+                    }
+
+                    cleanupOnDestroy(machine, message);
+                } catch (NotFoundException | MachineException ignore) {
+                }
```
skabashnyuk: Log.warn ?

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
garagatyi: It uses private method, so I make it private class.

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
garagatyi: It uses private method of outer class

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
garagatyi: I would say it is normal situation. Can happen when user destroy machine by API

skabashnyuk:
other ok

PR #288, opened 2015-09-25
IDEX-2990 tree bugfixes and improvements

vzhukovskii:
@vparfonov @skabashnyuk 

- added "scroll from source" button, like in IDEA
- fixed refreshing project explorer after creating module
- refactor delete item action to allow work through promises
- added ability to handle delete key
- refactored delete dialog to show additional information about deleted items like in IDEA
- added additional validation to item rename, so that item cannot be renamed into self name
- improved project explorer refreshing files mechanism after renaming folder with similar name
- added additional event SynchronizeProjectViewEvent to allow reload files throgh eventbus
- fix selection changing after right click on the node
- move NewProjectExplorer* names into persisted ProjectExplorer* names


skabashnyuk:
ok for me

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/project/node/remove/DeleteNodeHandler.java
```diff
@@ -0,0 +1,408 @@
+/*******************************************************************************
```
vparfonov: No test?

vzhukovskii: At this moment I blocked with creating test for this class because methods bellow are use static methods for Promises. And mockito testing library doesn't support mock static methods (http://stackoverflow.com/a/4482364). We may use PowerMock for testing or refactor Promises class to extract interface from it and make it injectable to allow writing tests for other stuff.

PR #292, opened 2015-09-29
IDEX-2985: Added three points when project name is too long in the pr…

AndrienkoAleksandr:
Added three points when project name is too long in the project explorer


vzhukovskii:
ok

skabashnyuk:
+0

PR #295, opened 2015-09-29
IDEX-3085. Recognize project types after import raw project sources and make user possible to choose desired one.

RomanNikitenko:
Recognize project types after import raw project sources and make user possible to choose desired one.


skabashnyuk:
LGTM

vparfonov:
ok

PR #299, opened 2015-10-01
IDEX-2603 All DTO objects and JsonSerializable will be serialized in the same way.

skabashnyuk:
All DTO objects and JsonSerializable will be serialized in the same way

Some background.
Server side DTO do not have @Dto annotation. This annotation is located in the interface which implements this particular class.
Server DTO classes has annotation @org.eclipse.che.dto.shared.DTOImpl and they are implements JsonSerializable interface.
Behind the scene DtoFactory.getInstance().toJson(t) use JsonSerializable to convert object to json.
To make sure that all objects serialized in the same thay I've reworked  codition.
So from now we check if object implement JsonSerializable interface instead of checking of DTO annotation.

Signed-off-by: Sergii Kabashniuk <skabashnyuk@codenvy.com>


garagatyi:
ok

PR #307, opened 2015-10-02
Idex 2812

evoevodin:
Decouple workspace from account. 

Add WorkspaceHooks implemenation for account which allows or rejects workspace start.


evoevodin:
@garagatyi, @akorneta, @skabashnyuk please review

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountService.java
```diff
@@ -120,8 +130,8 @@ public AccountService(AccountDao accountDao,
     @POST
     @GenerateLink(rel = Constants.LINK_REL_CREATE_ACCOUNT)
     @RolesAllowed({"user", "system/admin"})
-    @Consumes(MediaType.APPLICATION_JSON)
-    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
```
garagatyi: I didn't think about static import in such cases before.
idea :+1:

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountService.java
```diff
@@ -580,14 +588,49 @@ public void remove(@ApiParam(value = "Account ID", required = true)
...
 
+    @POST
+    @Path("/{accountId}/{workspaceId}")
+    @Produces(APPLICATION_JSON)
+    @RolesAllowed("account/owner")
+    public AccountDescriptor registerWorkspace(@PathParam("accountId") String accountId, @PathParam("workspaceId") String workspaceId)
+            throws NotFoundException, ServerException, BadRequestException, ConflictException {
+        Account account = accountDao.getById(accountId);
+        UsersWorkspace workspace = workspaceManager.getWorkspace(workspaceId);
+        if (accountDao.isWorkspaceRegistered(workspaceId)) {
```
garagatyi: Maybe change method to PUT and allow register it multiple times?

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/shared/dto/AccountDescriptor.java
```diff
@@ -44,6 +46,12 @@
 
     AccountDescriptor withId(String id);
 
+    List<UsersWorkspaceDto> getWorkspaces();
```
garagatyi: What if we extent Account with this DTO?

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -488,68 +460,29 @@ private MachineDto asDto(Machine machine) {
...
                                        .withLinks(singletonList(machineLink));
     }
 
-    private MachineSourceDto asDto(MachineSource source) {
+    private MachineSourceDto asRuntimeWorkspaceDto(MachineSource source) {
         return newDto(MachineSourceDto.class).withType(source.getType()).withLocation(source.getLocation());
     }
 
-    private ServerDto asDto(Server machine) {
+    private ServerDto asRuntimeWorkspaceDto(Server machine) {
```
garagatyi: Strange method name for functionality method provides

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -488,68 +460,29 @@ private MachineDto asDto(Machine machine) {
...
                                        .withProperties(machine.getProperties())
-                                       .withSource(asDto(machine.getSource()))
+                                       .withSource(asRuntimeWorkspaceDto(machine.getSource()))
                                        .withServers(servers)
                                        .withMetadata(asDto(machine.getMetadata()))
                                        .withLinks(singletonList(machineLink));
     }
 
-    private MachineSourceDto asDto(MachineSource source) {
+    private MachineSourceDto asRuntimeWorkspaceDto(MachineSource source) {
```
garagatyi: Strange method name

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -470,13 +442,13 @@ private RuntimeWorkspaceDto asDto(RuntimeWorkspace workspace) {
...
                                                 .withActiveEnvName(workspace.getActiveEnvName())
-                                                .withDevMachine(asDto(workspace.getDevMachine()))
+                                                .withDevMachine(asRuntimeWorkspaceDto(workspace.getDevMachine()))
                                                 .withRootFolder(workspace.getRootFolder())
                                                 .withMachines(machines)
                                                 .withLinks(links);
     }
 
-    private MachineDto asDto(Machine machine) {
+    private MachineDto asRuntimeWorkspaceDto(Machine machine) {
```
garagatyi: Strange method name

platform-api/che-core-api-account/pom.xml
```diff
@@ -39,15 +39,16 @@
...
-
-
-
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-workspace</artifactId>
             <version>${project.version}</version>
         </dependency>
-
+        <dependency>
```
garagatyi: I didn't find usage of that library in this PR. Can you point me out?

platform-api/che-core-api-account/pom.xml
```diff
@@ -39,15 +39,16 @@
...
+            <groupId>org.modelmapper</groupId>
+            <artifactId>modelmapper</artifactId>
+            <version>0.7.5</version>
```
garagatyi: Move version to dependency management

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
evoevodin: Yes i agree, i missed this refactoring effect

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/shared/dto/AccountDescriptor.java
evoevodin: Account is a class and i don't want to separate it on interface and impl, at least not now

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -190,7 +182,7 @@ public UsersWorkspaceDto getByName(@PathParam("name") String name) throws Server
         //TODO add maxItems & skipCount to manager
         return workspaceManager.getRuntimeWorkspaces(getCurrentUserId())
                                .stream()
-                               .map(this::asDto)
+                               .map(this::asRuntimeWorkspaceDto)
```
garagatyi: Can you use DtoConverter here?

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -199,14 +191,14 @@ public UsersWorkspaceDto getByName(@PathParam("name") String name) throws Server
     @Produces(APPLICATION_JSON)
     public RuntimeWorkspaceDto getRuntimeWorkspaceById(@PathParam("id") String id)
             throws ServerException, BadRequestException, NotFoundException {
-        return asDto(workspaceManager.getRuntimeWorkspace(id));
+        return asRuntimeWorkspaceDto(workspaceManager.getRuntimeWorkspace(id));
```
garagatyi: Can you use DtoConverter here?

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -199,14 +191,14 @@ public UsersWorkspaceDto getByName(@PathParam("name") String name) throws Server
...
     }
 
     @GET
     @Path("/name/{name}/runtime")
     @Produces(APPLICATION_JSON)
     public RuntimeWorkspace getRuntimeWorkspaceByName(@PathParam("name") String name) {
-        return asDto(workspaceManager.getRuntimeWorkspace(name, getCurrentUserId()));
+        return asRuntimeWorkspaceDto(workspaceManager.getRuntimeWorkspace(name, getCurrentUserId()));
```
garagatyi: Can you use DtoConverter here?

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -392,35 +387,12 @@ private UsersWorkspaceDto asUsersWorkspaceDto(UsersWorkspace workspace) {
...
-                                              .withCommands(commands)
-                                              .withProjects(projects)
-                                              .withEnvironments(environments)
-                                              .withAttributes(workspace.getAttributes())
-                                              .withLinks(links);
+        return workspace.withLinks(links);
     }
 
-    private RuntimeWorkspaceDto asDto(RuntimeWorkspace workspace) {
+    private RuntimeWorkspaceDto asRuntimeWorkspaceDto(RuntimeWorkspace workspace) {
```
garagatyi: Can you replace it with DtoConverter and injectLinks method?

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
garagatyi: Can you replace it with DtoConverter?

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountWorkspaceHooks.java
```diff
@@ -0,0 +1,143 @@
...
+                                                                                                  ForbiddenException,
+                                                                                                  ServerException {
+        User currentUser = EnvironmentContext.getCurrent().getUser();
+        if (accountId == null) {
+            // check if account is already specified for given workspace
+            if (!accountDao.isWorkspaceRegistered(workspace.getId())) {
+                // when workspace is not added to any of accounts
+                // try to detect whether current user is owner of a single account
+                List<Account> ownedAccounts = accountDao.getByOwner(currentUser.getId());
+                // there is no way to get empty list, user always has at least one account
```
garagatyi: Are you sure that situation when user have no account is not possible?

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountWorkspaceHooks.java
```diff
@@ -0,0 +1,143 @@
...
+                    throw new ForbiddenException(format("Workspace start rejected. Impossible to determine account for workspace '%s', "
+                                                        + "user '%s' is owner of several accounts. Specify account identifier!",
+                                                        workspace.getId(),
+                                                        workspace.getOwner()));
+                }
+                // account detection is completed
+                // if user is owner of a single account add workspace to it
+                Account account = ownedAccounts.get(0);
+                account.getWorkspaces().add(workspace);
+                accountDao.update(account);
```
garagatyi: This is very implicit side effect. I don't like it

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountWorkspaceHooks.java
```diff
@@ -0,0 +1,143 @@
...
+                              .noneMatch(member -> currentUser.getId().equals(member.getUserId())
+                                                   && member.getRoles().contains("account/owner"))) {
+                    throw new ForbiddenException(format("Workspace start rejected. User '%s' doesn't own account '%s'",
+                                                        currentUser.getId(),
+                                                        accountId));
+                }
+                // otherwise when user is owner of specified account add workspace to it
+                Account account = accountDao.getById(accountId);
+                account.getWorkspaces().add(workspace);
+                accountDao.update(account);
```
garagatyi: Implicit side effect

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountService.java
evoevodin: The concept is close to [this one](https://github.com/codenvy/che-core/blob/4.0/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java#L244-L256).
I think `PUT` may be used in this case, but `POST` looks okay too. Maybe guys have something to say about it @skabashnyuk, @akorneta ?

The workspace relation to account is _many to one_, it is important because in the other way it would not be possible to detect the account which should pay for used resources. Or what do you mean by multiple times?

garagatyi: I mean method PUT is idempotent

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/shared/dto/AccountDescriptor.java
garagatyi: OK

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
evoevodin: I decided to move to `DtoConverter` only methods which are need for `AccountService`. 
It is kind of temporary solution which will be replaced with _Registry_ for DTO transformers [asap](https://github.com/codenvy/che-core/blob/IDEX-2812/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DtoConverter.java#L34).
I think [model mapper](http://modelmapper.org/) is good enough for it, but it worth some time(which i don't have) to investigate it.

platform-api/che-core-api-account/pom.xml
evoevodin: Oops, it is not used, i just forgot to remove it. Thank you!

platform-api/che-core-api-account/pom.xml
evoevodin: The same [comment](https://github.com/codenvy/che-core/pull/307#discussion_r41075416)

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
evoevodin: The same [comment](https://github.com/codenvy/che-core/pull/307#discussion_r41075354)

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
evoevodin: The same [https://github.com/codenvy/che-core/pull/307#discussion_r41075354](comment)

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
evoevodin: The same [comment](https://github.com/codenvy/che-core/pull/307#discussion_r41075354)

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
evoevodin: The same [comment](https://github.com/codenvy/che-core/pull/307#discussion_r41075354)

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountWorkspaceHooks.java
evoevodin: Actually i was sure about it.
Do you have an example when it is possible?

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountWorkspaceHooks.java
evoevodin: As far as i remember we decided to use both implicit and explicit approaches.

_The point for implicit:_ in 90% situations user is account/owner of a single account, then it is possible to detect it and use, which simplifies the usage of the API.
_The point for explicit:_ when user explicitly specifies workspace it is obvious and has no side effects.

Actually i agree with you and i think that workspace registration should be done in explicit way.

Btw all cases are described [here](https://github.com/codenvy/che-core/blob/IDEX-2812/platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountWorkspaceHooks.java#L57-L70)

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountWorkspaceHooks.java
garagatyi: It was just suggestion. For example if user was just created via API.

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountService.java
akorneta: +1 to POST

platform-api-client-gwt/che-core-client-gwt-account/pom.xml
```diff
@@ -102,6 +107,7 @@
                 <configuration>
                     <dtoPackages>
                         <package>org.eclipse.che.api.account.shared.dto</package>
+                        <package>org.eclipse.che.api.workspace.shared.dto</package>
```
sleshchenko: Why do we need generate implementations for workspace dao here?

evoevodin: If you mean workspace *dto*, it doesn't work without it. 
`Account` dependes on `UsersWorkspace` it  means  that `AccountDescriptor` needs `UsersWorkspaceDto` which is defined in `org.eclipse.che.api.workspace.shared.dto`.
It would not compile if you remove this definition. 

Do you have an another solution?

sleshchenko: OK, thanks. You're right. It is not good but I don't have another solution.

PR #309, opened 2015-10-05
IDEX-2965: removed the ability to import project without type and spe…

akorneta:
…cified configs

### DIFF
--- a/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
+++ b/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
@@ -18,6 +18,7 @@ import com.wordnik.swagger.annotations.ApiResponse;
 import com.wordnik.swagger.annotations.ApiResponses;
 
 import org.apache.commons.fileupload.FileItem;
+import org.eclipse.che.api.core.BadRequestException;
 import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.ForbiddenException;
 import org.eclipse.che.api.core.NotFoundException;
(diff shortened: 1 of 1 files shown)


platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -811,55 +812,52 @@ private VirtualFileEntry getVirtualFile(String workspace, String path, boolean f
...
-            if (newProject != null) {
-                visibility = newProject.getVisibility();
-                RunnersDescriptor descriptor = newProject.getRunners();
-                if (descriptor != null) {
-                    replaceDefaultRunnerPath(descriptor);
-                }
-                projectConfig = DtoConverter.fromDto2(newProject, projectTypeRegistry);
-            }
+            visibility = newProject.getVisibility();
+            RunnersDescriptor descriptor = newProject.getRunners();
```
mshaposhnik: Never can be null?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -874,6 +872,13 @@ private ImportResponse configureProject(ImportProject importProject, FolderEntry
                 postImportProjectHandler.onProjectImported(project.getBaseFolder());
             }
         } catch (ConflictException | ForbiddenException | ServerException | NotFoundException e) {
+            if (newProject.getDescription() != null
+                || newProject.getAttributes() != null
+                || newProject.getRunners() != null
+                || newProject.getBuilders() != null
+                || newProject.getMixinTypes() != null) {
+                throw new BadRequestException("Impossible to save configurations for the project without type");
```
mshaposhnik: this code looks strange. how the server exception related to the given message ?

garagatyi:
LGTM
BTW code in ProjectService is completely unclear. Code reviews should not accept such code, but I suppose it was done even before we started make code reviews by multiple people in that area of project.

skabashnyuk:
LGTM

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
akorneta: if it be null then this method throw npe here https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java#L888

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
akorneta: yes its looks strange. But it's occurs when we try convert dto into project's configs.

PR #323, opened 2015-10-08
IDEX-2899. Remove module by absolute and relative path on server side.

RomanNikitenko:
Remove module by absolute and relative path on server side.


platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/handlers/RemoveModuleHandler.java
```diff
@@ -0,0 +1,26 @@
...
+ *******************************************************************************/
+package org.eclipse.che.api.project.server.handlers;
+
+import org.eclipse.che.api.core.ConflictException;
+import org.eclipse.che.api.core.ForbiddenException;
+import org.eclipse.che.api.core.ServerException;
+import org.eclipse.che.api.project.server.FolderEntry;
+import org.eclipse.che.api.project.server.ProjectConfig;
+
+/**
```
vparfonov: javadoc

PR #326, opened 2015-10-09
Fix API methods for working with oauth2 token

sleshchenko:
Add method for retrieving OAuth token
Fix NPE in OAuthAuthenticator
Add RolesAllowed annotation to invalidate method
Reformat the code

### DIFF
--- a/platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
+++ b/platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
@@ -10,23 +10,30 @@
  *******************************************************************************/
 package org.eclipse.che.security.oauth;
 
-import com.google.common.base.Strings;
-
+import org.eclipse.che.api.auth.shared.dto.OAuthToken;
+import org.eclipse.che.api.core.BadRequestException;
+import org.eclipse.che.api.core.ForbiddenException;
+import org.eclipse.che.api.core.NotFoundException;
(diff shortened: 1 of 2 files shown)


platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
```diff
@@ -147,13 +152,56 @@ public Response callback() throws OAuthAuthenticationException {
...
+    @GET
+    @Path("/token/invalidate")
+    @RolesAllowed({"user", "temp_user"})
+    public Response invalidate(@Required @QueryParam("oauth_provider") String oauthProvider) throws IOException, BadRequestException {
+        OAuthAuthenticator oauth = getAuthenticator(oauthProvider);
+        final Principal principal = security.getUserPrincipal();
+        if (principal != null && oauth.invalidateToken(principal.getName())) {
+            return Response.ok().build();
+        }
+        return Response.status(404)
```
garagatyi: Why not throw NotFoundException here and set method return type to void

garagatyi:
LGTM

platform-api/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
sleshchenko: Nice remark. Thanks

skabashnyuk:
LGTM

PR #330, opened 2015-10-09
Idex 3076

vitaliy-guliy:


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/widgets/listtab/ListButtonWidget.java
```diff
@@ -29,106 +31,112 @@
...
-
-        addDomHandler(this, ClickEvent.getType());
-
-        this.items = new ArrayList<>();
+        closeTime = System.currentTimeMillis();
+        addDomHandler(new MouseDownHandler() {
+            @Override
+            public void onMouseDown(MouseDownEvent event) {
+                long time = System.currentTimeMillis();
+                if (time - closeTime < 100) {
```
vzhukovskii: some magic?

vitaliy-guliy: The popup panel is autoclosed and I had no idea how to prevent opening it after pressing menu button. So, decided to use time to detect the action.

PR #336, opened 2015-10-14
IDEX-3241 Server side API allowing to manage Che extensions + plugins

benoitf:
Change-Id: I6922c0d8be1a3d0f5c9ceb930326b52cb5b096b6


che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/ChePluginConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+        return new File(System.getProperty("catalina.base")).toPath();
+    }
+
+    /**
+     * @return local path to maven repository
+     */
+    @Override
+    public Path getLocalMavenRepository() {
+        Path userHome = Paths.get(System.getProperty("user.home"));
+        Path m2Repository = userHome.resolve(".m2/repository");
```
skabashnyuk: Are you sure this is true for all places
For example I can override it in settings.xml

che-admin-api/che-core-api-plugin/pom.xml
```diff
@@ -0,0 +1,206 @@
...
+    <packaging>jar</packaging>
+    <name>Che Core :: Admin API :: Plugin</name>
+    <properties>
+        <dto-generator-out-directory>${project.build.directory}/generated-sources/dto/</dto-generator-out-directory>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-aether-provider</artifactId>
+            <version>3.3.1</version>
```
skabashnyuk: is this compile or build dependency?

che-admin-api/che-core-api-plugin/pom.xml
```diff
@@ -0,0 +1,206 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.aether</groupId>
+            <artifactId>aether-api</artifactId>
+            <version>1.0.2.v20150114</version>
```
skabashnyuk: version should be taken from depmgt pom

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/CustomExceptionMapper.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * Customize the exceptions to bring some HTTP status
+ * @author Florent Benoit
+ */
+@Provider
+@Singleton
+public class CustomExceptionMapper implements ExceptionMapper<PluginException> {
```
skabashnyuk: Can you use existed ApiExceptions ?

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/PluginsService.java
```diff
@@ -0,0 +1,233 @@
...
+    public PluginInstall newInstall() throws PluginManagerException, PluginInstallerException {
+        return dtoBuilder.convert(pluginManager.requireNewInstall());
+    }
+
+    /**
+     * Gets the list of all install
+     * @return the IDs of every install
+     */
+    @GET
+    @Path("/install")
```
skabashnyuk: Path "install" can be confusing.
Install means change something.
 May be name "list" is better?

che-admin-api/che-core-api-plugin/pom.xml
benoitf: Runtime

che-admin-api/che-core-api-plugin/pom.xml
benoitf: It's OK on these dependencies ?

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/ChePluginConfiguration.java
benoitf: yes you can ;-) 
I may add extra settings parsing. Btw it's to help to download remote artifacts with maven/ aether so it's not a problem

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/PluginsService.java
benoitf: It has run mvn install to generate war file

che-admin-api/che-core-api-plugin/pom.xml
skabashnyuk: it will be packaged with Che distribution or part of some artifact?

che-admin-api/che-core-api-plugin/pom.xml
skabashnyuk: if they are from Eclipse yes.

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/PluginsService.java
skabashnyuk: sorry?

benoitf: it's to list all installations that have been done.
Like listing all processes, etc
It's not a verb

because we have /install and /install/<id>
like we could have /dogs/ and /dogs/<id>

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/CustomExceptionMapper.java
benoitf: We're forced to use some Che API exception class on the mapping or we're not allowed to use inner exceptions ?
I mean, rethrowing internal exception to existing API exception or throwing internally the API exceptions ?

che-admin-api/che-core-api-plugin/pom.xml
benoitf: As part of the Che Admin war file only packaged alongside Eclipse Che for now.
(It's not inside che.war as we need to be able to administrate it)

benoitf:
Use of versions from che-depmgt

che-admin-api/che-core-api-plugin/pom.xml
```diff
@@ -0,0 +1,206 @@
...
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>che-admin-api-parent</artifactId>
+        <groupId>org.eclipse.che.core</groupId>
+        <version>3.13.1-SNAPSHOT</version>
+    </parent>
+    <artifactId>che-core-api-plugin</artifactId>
```
skabashnyuk: I have doubts about project name and location
@gazarenkov  can you comment this. What is prefered name for this project

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/PluginsService.java
```diff
@@ -0,0 +1,233 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
+import java.io.File;
+import java.util.List;
+
+/**
+ * JAX-RS Plugins service allowing to list, install, uninstall plugins, etc.
+ * @author Florent Benoit
+ */
```
skabashnyuk: Is this service will be used in che only? Not in hosted version

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/internal/api/PluginManager.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.internal.api;
+
+import java.util.List;
+
+/**
+ * @author Florent Benoit
```
skabashnyuk: javadoc

che-admin-api/che-core-api-plugin/src/main/java/org/eclipse/che/plugin/PluginsService.java
benoitf: yes not in hosted

skabashnyuk: then may be it should be moved to che. WDYT?

benoitf: it may be used on OnPrem

skabashnyuk: how methods will be protect? This class made potentially risky operations.

skabashnyuk:
LGTM you can merge it in master

benoitf:
PR has been moved to che repository

PR #339, opened 2015-10-16
IDEX-3267 Return any-any permissions for all VFS entries

skabashnyuk:
In 4.0 we are not relying on vfs permission so we can simply return any-any permission
for all users. We need to return something because UI is still relying on this information.

### DIFF
--- a/che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
+++ b/che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
@@ -11,7 +11,6 @@
 package org.eclipse.che.vfs.impl.fs;
 
 import com.google.common.annotations.Beta;
-import com.google.common.collect.Sets;
 import com.google.common.hash.HashFunction;
 import com.google.common.hash.Hashing;
 import com.google.common.io.ByteSource;
@@ -47,7 +46,6 @@ import org.eclipse.che.api.vfs.shared.PropertyFilter;
 import org.eclipse.che.api.vfs.shared.dto.AccessControlEntry;
 import org.eclipse.che.api.vfs.shared.dto.Principal;
 import org.eclipse.che.api.vfs.shared.dto.Property;
-import org.eclipse.che.api.vfs.shared.dto.VirtualFileSystemInfo;
 import org.eclipse.che.api.vfs.shared.dto.VirtualFileSystemInfo.BasicPermissions;
 import org.eclipse.che.commons.lang.NameGenerator;
 import org.eclipse.che.commons.lang.Pair;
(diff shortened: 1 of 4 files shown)


che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
```diff
@@ -296,6 +250,10 @@ protected AccessControlList loadValue(Path key) {
 
     private final VirtualFileSystemUserContext userContext;
 
+
+
+    private final AccessControlList DEFAULT_ACL = new AccessControlList();
```
evoevodin: It is not a constant, so it should not be defined in `CONSTANT_CASE`

che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
```diff
@@ -323,11 +278,27 @@ protected AccessControlList loadValue(Path key) {
...
+
+        acl.add(newDto(AccessControlEntry.class)
+                        .withPrincipal(newDto(Principal.class)
+                                               .withName("temp_user")
+                                               .withType(Principal.Type.GROUP))
+                        .withPermissions(singletonList("all")));
+        DEFAULT_ACL.update(acl, true);
+
+
+
```
evoevodin: Why there are 4 empty lines?

che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
```diff
@@ -297,6 +249,11 @@ protected AccessControlList loadValue(Path key) {
     private final VirtualFileSystemUserContext userContext;
 
     /**
+     * ACL that returned for any fs entry.
+     */
+    private final AccessControlList defaultAcl = new AccessControlList();
```
benoitf: I would say that it needs to be static as well as it's immutable and not defined per instance

skabashnyuk: As we expect in near future to remove this code at all I would say such complexity is not necessary. The purpose of this change is to make things working until we remove VFS at all.

benoitf: so it needs to introduce @deprecated stuff as well

skabashnyuk: Module marked as deprecated

benoitf:
+1

PR #341, opened 2015-10-19
CLDIDE-2569: add runner validator remove unnessesary code

akorneta:


platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -1533,4 +1527,55 @@ private String projectPath(String path) {
...
+
+    /**
+     * Checks the validity of project name.
+     *
+     * @param name
+     *         project name
+     * @throws BadRequestException
+     *         when if the project name is null or empty of is invalid
+     */
+    void isValidProjectName(String name) throws BadRequestException {
```
evoevodin: It is a bit strange to see the method which starts from `is` and returns nothing.
I this case you should use `check` method instead of `is`

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -1533,4 +1527,55 @@ private String projectPath(String path) {
...
+
+    /**
+     * Checks the validity of project runners.
+     *
+     * @param runnersDescriptor
+     *         project runners
+     * @throws BadRequestException
+     *         when runner name empty or is invalid
+     */
+    void isValidProjectRunners(RunnersDescriptor runnersDescriptor) throws BadRequestException {
```
evoevodin: I have already seen this above `validateProjectRunnerNames`.
It duplicates the same logic + uses `Pattern.compile` for each runner name.

akorneta: :+1:

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -852,13 +860,13 @@ private ImportResponse configureProject(ImportProject importProject, FolderEntry
                                          null,
                                          visibility);
                 RunnersDescriptor projectModuleRunners = projectModule.getRunners();
-                if (projectModuleRunners != null
-                    && projectModuleRunners.getDefault() != null
-                    && projectModuleRunners.getDefault().startsWith("project:/")) {
-                    replaceDefaultRunnerPath(projectModuleRunners);
+                if (projectModuleRunners != null && projectModuleRunners.getDefault() != null) {
+                    projectModuleRunners.getDefault();
```
sleshchenko: Why do you call it here and ignore result?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -1285,9 +1279,11 @@ public ItemReference getItem(@ApiParam(value = "Workspace ID", required = true)
         final List<TreeElement> nodes = new ArrayList<>(children.size());
         for (VirtualFileEntry child : children) {
         	if (child.isFolder()) {
-        		nodes.add(dtoFactory.createDto(TreeElement.class).withNode(DtoConverter.toItemReferenceDto((FolderEntry)child, uriBuilder.clone())).withChildren(getTree((FolderEntry)child, depth - 1, includeFiles, uriBuilder, dtoFactory)));
+        		nodes.add(dtoFactory.createDto(TreeElement.class).withNode(DtoConverter.toItemReferenceDto((FolderEntry)child, uriBuilder.clone())).withChildren(
```
sleshchenko: I think will be better to write something like
> nodes.add(dtoFactory.createDto(TreeElement.class)
>                            .withNode(DtoConverter.toItemReferenceDto((FolderEntry)child, uriBuilder.clone()))
>                            .withChildren(getTree((FolderEntry)child, depth - 1, includeFiles, uriBuilder, dtoFactory)));

skabashnyuk:
LGTM

PR #343, opened 2015-10-19
IDEX-3240 Move from che/ repository the maven plugin

benoitf:
Change-Id: Icb9e4fe5453200e085b41af4ad56f0e468132368


skabashnyuk:
LGTM

PR #346, opened 2015-10-21
IDEX-3300: Add check that workspace belongs to account owner

vinokurig:
Add account query parameter to getAllRunningProcesses API method.
It is needed to get all running processes that related to account.
For now if we get all running processes we will see processes that related to current user, but if there is a run process started from temporary workspace we will not see it.

### DIFF
--- a/platform-api/che-core-api-runner/src/main/java/org/eclipse/che/api/runner/RunnerService.java
+++ b/platform-api/che-core-api-runner/src/main/java/org/eclipse/che/api/runner/RunnerService.java
@@ -16,7 +16,11 @@ import com.wordnik.swagger.annotations.ApiParam;
 import com.wordnik.swagger.annotations.ApiResponse;
 import com.wordnik.swagger.annotations.ApiResponses;
 
+import org.eclipse.che.api.account.server.dao.AccountDao;
+import org.eclipse.che.api.account.server.dao.Member;
+import org.eclipse.che.api.core.ForbiddenException;
 import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.rest.HttpJsonHelper;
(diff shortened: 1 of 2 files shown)


platform-api/che-core-api-runner/src/test/java/org/eclipse/che/api/runner/RunnerServiceTest.java
```diff
@@ -283,29 +294,78 @@ public void testGetRunningProcesses() throws Exception {
...
+        processes = service.getRunningProcesses(null, "project", null);
         assertEquals(processes.size(), 0);
 
         // no project with that workspace + name
-        processes = service.getRunningProcesses("dummy", "project");
+        processes = service.getRunningProcesses("dummy", "project", null);
         assertEquals(processes.size(), 0);
     }
 
+    /**
```
sleshchenko: I think javadocs is redundant in tests. Except private method and specific cases

platform-api/che-core-api-runner/src/main/java/org/eclipse/che/api/runner/RunnerService.java
```diff
@@ -149,55 +167,82 @@ public ApplicationProcessDescriptor getStatus(@ApiParam(value = "Workspace ID",
...
         }
         return processes;
     }
 
+    private boolean isRelatedToAccountOrUser(RunQueueTask task, String accountId, String userId) throws ServerException {
+        if (accountId != null) {
+            try {
+                return accountId.equals(workspaceDao.getById(task.getRequest().getWorkspace()).getAccountId());
+            } catch (NotFoundException e) {
+                //Workspace does not exist
```
sleshchenko: 'Workspace does not exist' is not informative. You should write something about 'How we respond on this situation'

sleshchenko:
Other OK for me. @skabashnyuk WDYT?

skabashnyuk:
LGTM

PR #348, opened 2015-10-21
Skip editor close when node removed

vzhukovskii:
Add ability to configure editor behaviour
in case when we try to update project tree.
By default editor tries to close opened file
if it was removed after reloading. Now it
become possible to configure this behaviour.
This fix needed for java rename refactoring
feature.

@vparfonov

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/SearchNodeHandler.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/SearchNodeHandler.java
@@ -48,6 +48,7 @@ public class SearchNodeHandler implements ExpandNodeHandler, BeforeExpandNodeHan
     private AsyncCallback<Node> callback;
 
     private boolean forceUpdate = false;
+    private boolean closeMissingFiles = true;
 
     public SearchNodeHandler(Tree tree) {
         this.tree = tree;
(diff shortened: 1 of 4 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/SearchNodeHandler.java
```diff
@@ -222,6 +233,10 @@ public boolean isInSearchMode() {
         return inSearchMode;
     }
 
+    public boolean isCloseMissingFiles() {
+        return closeMissingFiles;
```
skabashnyuk: javadoc

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/SearchNodeHandler.java
```diff
@@ -58,8 +59,18 @@ public SearchNodeHandler(Tree tree) {
...
+    /**
+     * Search node in the project explorer tree by storable path.
+     *
+     * @param path path to node
+     * @param forceUpdate force children reload
+     * @param closeMissingFiles allow editor to close removed files if they were opened
+     * @return
+     * promise object with found node or promise error if node wasn't found
+     */
+    public Promise<Node> getNodeByPath(final HasStorablePath path, boolean forceUpdate, boolean closeMissingFiles) {
```
skabashnyuk: is this method do simple search without modification or.
Its something like: getNodeByPathAndDoSomeModificationDependsOnInputParameter?

vzhukovskii: closeMissingFiles parameter holds simple flag to pass it into project explorer module after node found. Business logic on file closing need to be moved from project explorer module into search node handler, but it may took few hours. Thats why as temporary solutions I decided to do it in this way

evoevodin:
Javadocs are not formatted.

vzhukovskii:
@evoevodin fixed

vparfonov:
ok

PR #349, opened 2015-10-21
Make all DTO-related projects modules of same sub-parent

tareqhs:
This allow for a better lifecycle management of all DTO stuff and paves the way for conveniently adding more tools that aid the DTO framework later on.


skabashnyuk:
Not clear for me what issue you are trying to solve. Can you clarify?

tareqhs:
Hi @skabashnyuk ,
First, right now the source of DTO is not grouped into a single hierarchy which makes it less convenient to develop and build it like other parts of Che code. Second, I'm planning to propose alternative tooling for DTO generation based on standard annotation processing instead of the current set of method, this might end up having three non-grouped DTO projects.

benoitf:
hi, about the new tooling, it will be proposed as a replacement of the current maven plugin ?

tareqhs:
hi @benoitf ,
the new tooling is a separate issue, but yes - i'll suggest gradually replacing the current method with a standard annotation processing one. another important thing i will introduce i using text templates instead of the unmanageable 'string append' hundreds of times all over the generator source code. I'll keep you posted about this.

benoitf:
ok, so +1 for me grouping all the dto modules into a dedicated maven module

skabashnyuk:
@tareqhs I don't see a big profit of grouping two artifacts in some folder. Yes it can looks simple but didn't bring some huge profit. Is there any technical reason to do so?

tareqhs:
hi @skabashnyuk ,
the benefits we explained in the previous comment. gently asking, is "huge profit" a required criteria for patching Che? what about general clean-ups, refactoring, minor code alignments, ...

skabashnyuk:
I would like to avoid any artifact movement before you propose something concrete with new DTO generators.
BTW Since we have more then one maven plugin I'm start thinking about separate repository for maven staff. I think if you proposal with new DTO generator will become popular special "maven" git repository would be logical for this kind of staff.

PR #351, opened 2015-10-21
IDEX-3283: Add hotkey for switch tabs in editors

AndrienkoAleksandr:


commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/util/input/CharCodeWithModifiers.java
```diff
@@ -64,4 +64,24 @@ public static int computeKeyDigest(int modifiers, int charCode) {
...
+    /** {@inheritDoc} */
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        CharCodeWithModifiers that = (CharCodeWithModifiers)o;
+
+        if (modifiers != that.modifiers) return false;
+        return charCode == that.charCode;
```
sleshchenko: I think it will be better to write 
return modifiers == that.modifiers
          && charCode == that.charCode;

vitaliy-guliy: Not sure we need this code

commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/util/input/CharCodeWithModifiers.java
```diff
@@ -64,4 +64,24 @@ public static int computeKeyDigest(int modifiers, int charCode) {
     public static int computeKeyDigest(SignalEvent event) {
         return computeKeyDigest(ModifierKeys.computeModifiers(event), KeyCodeMap.getKeyFromEvent(event));
     }
+
+    /** {@inheritDoc} */
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
sleshchenko: Maybe it will better to use "instead of" instead of comparing of classes

commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/util/input/CharCodeWithModifiers.java
```diff
@@ -64,4 +64,24 @@ public static int computeKeyDigest(int modifiers, int charCode) {
...
+        return charCode == that.charCode;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public int hashCode() {
+        int result = modifiers;
+        result = 31 * result + charCode;
+        return result;
+    }
```
benoitf: as modifiers and charCode are final, I would avoid to compute hashCode everytime the method is invoked. hashcode result can be stored

skabashnyuk: This is an arguable way. Is there any reason to do so?

AndrienkoAleksandr:
Code was improved and refactored https://github.com/codenvy/che-core/pull/355

PR #353, opened 2015-10-21
Idex 3082 Manage the file’s resume info

vitaliy-guliy:


ide/che-core-ide-jseditor/src/main/java/org/eclipse/che/ide/jseditor/client/texteditor/EmbeddedTextEditorPartViewImpl.java
```diff
@@ -159,7 +159,7 @@ public void updateInfoPanelPosition(final TextPosition position) {
 
     @Override
     public void updateInfoPanelUnfocused(final int linecount) {
-        this.infoPanel.displayLineCount(linecount);
+//        this.infoPanel.displayLineCount(linecount);
```
skabashnyuk: Are you sure you need this code?

vitaliy-guliy: I'm not sure we will restore a part of functionality in the future. So, I decided to leave it commented.

PR #354, opened 2015-10-21
IDEX-3331 IDEX-3306: Fix unexpected closing files with the same name …

AndrienkoAleksandr:
Fix unexpected closing files with the same name after close only one project, fix posibility to open earlier opened files after reopen a project.


vparfonov:
ok

AndrienkoAleksandr:
Wrong number one of issue IDEX-3313

PR #359, opened 2015-10-22
IDEX-3131: Refactor machine model

garagatyi:


platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/workspace/EnvironmentState.java
```diff
@@ -0,0 +1,23 @@
...
+ *******************************************************************************/
+package org.eclipse.che.api.core.model.workspace;
+
+import org.eclipse.che.api.core.model.machine.MachineState;
+
+import java.util.List;
+
+/**
+ * @author Alexander Garagatyi
+ */
```
skabashnyuk: javadoc

skabashnyuk:
LGTM

PR #372, opened 2015-10-28
Rename API BranchCheckout to Checkout + add API support to checkout f…

freedomp:
…ile/s (+Native Git implementation)

Add Rebase command API
add Repository state to Status API (as String)
add support to set/get revision range (since, until) in Log API
-- Added above API changes to native git implementation, with empty implementation

Signed-off-by: Dror Cohen <dror.cohen@sap.com>

### DIFF
--- a/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
+++ b/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
@@ -14,7 +14,7 @@ import org.eclipse.che.api.core.ApiException;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.git.shared.AddRequest;
 import org.eclipse.che.api.git.shared.Branch;
-import org.eclipse.che.api.git.shared.BranchCheckoutRequest;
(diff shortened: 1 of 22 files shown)


skabashnyuk:
Eclipse CLA confirmed for dror.cohen@sap.com Eclipse CLA confirmed

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -111,12 +113,12 @@ public void add(AddRequest request) throws ApiException {
         }
     }
 
-    @Path("branch-checkout")
+    @Path("checkout")
```
mkuznyetsov: It needs to rename this path in GitServiceClientImpl too

skabashnyuk:
Could you please also rebase current master in your PR

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
freedomp: Added a commit which fixes the path in GitServiceClientImpl

PR #377, opened 2015-10-30
CLDIDE-2609: Removed test xml files with license issues

evoevodin:
3 xml files which have license issues were removed from test resources


skabashnyuk:
LGTM

PR #378, opened 2015-10-30
Cldide 2597 HttpJsonHelper improvements

skabashnyuk:
1. Removed user authorization token from url. For the cases when some component decided to put ulr in log file. This way we will remove possibility of token leaks.
2. We drop a hint to server that we expect application-json response. this needed if server have two method with different produces content type for same path.

### DIFF
--- a/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
+++ b/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package org.eclipse.che.api.core.rest;
 
+import com.google.common.io.CharStreams;
+
 import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.ForbiddenException;
 import org.eclipse.che.api.core.NotFoundException;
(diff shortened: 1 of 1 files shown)


platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
```diff
@@ -409,9 +409,9 @@ public String requestString(int timeout,
             final String authToken = getAuthenticationToken();
             if ((parameters != null && parameters.length > 0) || authToken != null) {
                 final UriBuilder ub = UriBuilder.fromUri(url);
-                if (authToken != null) {
-                    ub.queryParam("token", authToken);
-                }
+                //remove sensitive infromation from url.
```
evoevodin: typo

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
```diff
@@ -427,8 +427,13 @@ public String requestString(int timeout,
             conn.setReadTimeout(timeout > 0 ? timeout : 60000);
             try {
                 conn.setRequestMethod(method);
+                //drop a hint for server side that we want to receive application/json
+                conn.addRequestProperty(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON);
```
evoevodin: `HttpJsonHelper` is __JSON__ helper but still it contains `requestString` method, don't you think that this change may affect code which uses `requestString` method ?

skabashnyuk: No. because of this 
```java
if (!(contentType == null || contentType.startsWith(MediaType.APPLICATION_JSON))) {
    throw new IOException(conn.getResponseMessage());
}
```

evoevodin: Well, then it is okay, i missed this point

garagatyi:
LGTM

tal-sapan:
Hi,

This change removes the token parameter from the URL, but I don't see it added as a header. How does the token pass in the request after this change?

garagatyi:
@tal-sapan 
```
+                if (authToken != null) {
+                    conn.setRequestProperty(HttpHeaders.AUTHORIZATION, authToken);
+                }
```

tal-sapan:
@garagatyi 
Ok, thanks

PR #380, opened 2015-11-01
Use constants for names of .codenvy folder and core files inside of it

tareqhs:
Change-Id: Ia295b80af583aec957fbf462beec998ebeefced6


mkuznyetsov:
Merged as 176df0e6bed7e513f6bac8781b874b4f9de816f0

PR #381, opened 2015-11-01
Use a single source of truth for ignored virtual files

tareqhs:
Change-Id: Id92b40d3c7bb2909c7ca5f10afb69323d8afc3dc

### DIFF
--- a/platform-api/che-core-api-vfs/src/main/java/org/eclipse/che/api/vfs/server/util/VirtualFileDefaults.java
+++ b/platform-api/che-core-api-vfs/src/main/java/org/eclipse/che/api/vfs/server/util/VirtualFileDefaults.java
@@ -1,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2015 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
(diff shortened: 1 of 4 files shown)


platform-api/che-core-api-vfs/src/main/java/org/eclipse/che/api/vfs/server/util/VirtualFileDefaults.java
```diff
@@ -0,0 +1,36 @@
...
+     * Checks whether the given path should be ignored by the VFS implementations in filters and permission checks.
+     * 
+     * @param path
+     *            The file to check
+     * @return Whether should be ignored.
+     */
+    public static boolean isPathIgnored(Path path) {
+        int length = path.length();
+        // TODO use common constants for these
+        return length >= 2 && ".codenvy".equals(path.element(length - 2))
```
mkuznyetsov: Couldn't You use CODENVY_DIR constant for ".codenvy"? Looks like it is opposing to another PR: https://github.com/codenvy/che-core/pull/380

mkuznyetsov:
Merged as 4b33a524d9ae9a162190f1cf4dc3c19ec8817461

platform-api/che-core-api-vfs/src/main/java/org/eclipse/che/api/vfs/server/util/VirtualFileDefaults.java
tareqhs: Hi @mkuznyetsov ,
VFS projects and lower than project service and cannot use constants from it. A more general solution is to allow de-coupled injection of a 'master filter' that is implemented in project service, but this is a much bigger change. A first step is to simply use a single source of truth, later it can be further improved.

PR #383, opened 2015-11-02
IDEX-2470: Add checking of svn project type

vinokurig:
It is needed to send request to proper vcs service.


skabashnyuk:
This code needs deep refactoring. But in this cases we are not ready for it 3.x, So you can proceed and merge.

PR #386, opened 2015-11-04
Fix type of token in getToken method

sleshchenko:
@skabashnyuk


skabashnyuk:
Need more details.

sleshchenko:
getToken REST API method returns token as OAuthToken, not as String. But OAuthServiceClient expects token as sting. It is mistake

skabashnyuk:
LGTM

PR #389, opened 2015-11-04
IDEX-3049, IDEX-3319, IDEX-3449, IDEX-3450

vzhukovskii:
Tree changes:
- in node renderer changed manipulations with html, now work performs with dom nodes
- added possibility to customize node styles directly from node by css
- added method to refresh visual representation of specified node without reloading
- temporary disable inner checking for child in child, to proper rendering ">" symbol
- fixed setting parent after node intercepting, in some cases few nodes was without parent
- updated mechanism of firing changing selection node event

Upgraded behaviour when expanded node renames. Now expanded nodes saves their state
In some cases directory structure may be changed if renamed directory becomes source directory

Machie's project path requests only when ide starts to avoid unneccessary requests during selection change

Get project method in project service now returns Project Descriptos instead of Project Reference
This was need to avoid unneccessary requests to get project when we open it

Projects are always opened now. We can't close or open it. For now we can create or delete it

After init or delete git repository project explorer doesn't go crazy

@vparfonov


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/ProjectExplorerViewImpl.java
```diff
@@ -105,10 +111,18 @@
...
-    private ToolButton          scrollFromSourceButton;
-    private SearchNodeHandler   searchNodeHandler;
-    private HandlerRegistration closeEditorOnNodeRemovedHandler;
+    private ToolButton        goIntoBackButton;
+    private ToolButton        scrollFromSourceButton;
+    private ToolButton        refreshButton;
+    private ToolButton        collapseAllButton;
+    private SearchNodeHandler searchNodeHandler;
+
+    public static final String GO_INTO_BUTTON_ID            = "goBackButton";
```
sleshchenko: So, go into or go back?

vzhukovskii: go back, just copy/pase mistake

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/ProjectExplorerViewImpl.java
```diff
@@ -105,10 +111,18 @@
...
+    public static final String SCROLL_FROM_SOURCE_BUTTON_ID = "scrollFromSourceButton";
+    public static final String REFRESH_BUTTON_ID            = "refreshSelectedFolder";
+    public static final String COLLAPSE_ALL_BUTTON_ID       = "collapseAllButton";
+    public static final String PROJECT_TREE_WIDGET_ID       = "projectTree";
+    public static final String PROJECT_TREE_SCROLL_PANEL_ID = "projectExplorer";
```
sleshchenko: I think it is copy/paste error too

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/project/node/ProjectDescriptorNode.java
```diff
@@ -63,6 +58,13 @@ public void updatePresentation(@NotNull NodePresentation presentation) {
         presentation.setPresentableText(getData().getName());
         presentation.setPresentableIcon(isValid(getData()) ? nodeManager.getNodesResources().projectFolder()
                                                            : nodeManager.getNodesResources().notValidProjectFolder());
+        presentation.setPresentableTextCss("font-weight:bold");
+
+        if ("private".equals(getData().getVisibility())) {
```
sleshchenko: Do we show some icon If visibility equals to public? If no, why? I think it is not good behaviour to show icon for private project but not show some icon for public projects.

vzhukovskii: It's an experemental behaviour. If we'll decide to show some "unlock icon" for public projects it won't be difficult to add it.

sleshchenko: What issue for showing lock icon for private projects?

vzhukovskii: There isn't such issue.

sleshchenko: Who decided to show this icon?

vzhukovskii: I am.

sleshchenko: I disagree with you

ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/smartTree/TreeNodeLoader.java
```diff
@@ -128,10 +128,11 @@ public void onLoad(final LoadEvent event) {
                 }
             }
 
-            //Set joint element on the view
-            Promise<Void> promise = Promises.resolve(null);
-            List<Node> exist = tree.getNodeStorage().getChildren(event.getRequestedNode());
-            chainChildrenAndSetJointElement(promise, exist.listIterator());
+            //Temporary disable it
```
sleshchenko: Can you describe cause of disabling or when will be it enabled?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/project/node/ProjectDescriptorNode.java
vzhukovskii: I see. So we can delete this code which shows lock icon for private projects.

sleshchenko: Up to you

ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/smartTree/TreeNodeLoader.java
vzhukovskii: This method will be refactored with this issue https://jira.codenvycorp.com/browse/IDEX-3319
So to reduce requests count I temporary disable inner checking of childs to proper displayin ">" control of the node.

sleshchenko: I think you should write information like this in TODO comment

vzhukovskii: No problem/

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/ProjectExplorerViewImpl.java
```diff
@@ -555,16 +514,116 @@ public boolean isShowHiddenFiles() {
...
+
+        for (Map.Entry<String, Node> entry : oldIdToNode.entrySet()) {
+            if (!tree.getNodeStorage().reIndexNode(entry.getKey(), entry.getValue())) {
+                Log.info(getClass(), "Node wasn't re-indexed");
+            }
+        }
+
+        tree.refresh(node);
+    }
+
```
dimasnurenko: It seems that this method must be located on presenter. The view must be responsible only for view representation.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/ProjectExplorerViewImpl.java
```diff
@@ -555,16 +514,116 @@ public boolean isShowHiddenFiles() {
...
+        if (!event.getNode().isLeaf()) {
+            tree.getNodeLoader().loadChildren(event.getNode(), true);
+        }
+
+        //here we have possible odd behaviour: after renaming directory we should perform checking structure of the expanded directories
+        //cause some of them after renaming may become source directory and we need to replace them with correct nodes, possible solution
+        //is to use node interceptors to intercept expanded nodes
+    }
+
```
dimasnurenko: It seems that this method must be located on presenter. The view must be responsible only for view representation.
tree.getNodeStorage(),event.getNode() We can get this object once and use reference to it. I think this will be more readable.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/project/node/ProjectDescriptorNode.java
evidolob: I think we don't need privacy icon in 4.0. Vitalii

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/app/AppContext.java
```diff
@@ -107,28 +99,11 @@ public void setFactory(Factory factory) {
...
-    public void removeOpenedProject(ProjectDescriptor descriptor) {
-        openedProjects.remove(descriptor);
+    public String getProjectsRoot() {
+        return projectsRoot;
     }
 
-    /** Returns list of opened projects. */
-    public List<ProjectDescriptor> getOpenedProjects() {
-        return openedProjects;
+    public void setProjectsRoot(String projectsRoot) {
```
evidolob: Are you sure we need this logic in AppContext, i think it must be on Project layer

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/DefaultNodeInterceptor.java
```diff
@@ -32,7 +32,7 @@
     }
 
     @Override
-    public Integer weightOrder() {
-        return 0;
+    public int getPriority() {
```
evidolob: Need doc here

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/project/shared/DtoUtil.java
```diff
@@ -0,0 +1,70 @@
+/*******************************************************************************
```
evidolob: Not clear for me why we need this class

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/app/AppContext.java
vzhukovskii: This project root that comes from machine dto at extension server startup, this isn't a projects root related to any project in project explorer

vzhukovskii: This projects root comes when machines is starting. I don't think that we can move this onto project layer. So suitable place where I can store this property is in the app context.

vparfonov:
ok

PR #391, opened 2015-11-09
CLDIDE-2598: Use token in  RemoteTask & SourcesManagerImpl calls

evoevodin:
Each RemoteTask & SourcesManagerImpl request goes with authorization token which allows us to protect `SlaveBuilderService` from unauthorized calls.


skabashnyuk:
LGTM

garagatyi:
LGTM

PR #392, opened 2015-11-09
IDEX-3356: Change behavior import wizard

akorneta:


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/projectimport/wizard/ImportWizard.java
```diff
@@ -98,15 +95,19 @@ public void complete(@NotNull CompleteCallback callback) {
...
                 callback.onFailure(new Exception(localizationConstant.createProjectFromTemplateProjectExists(projectName)));
             }
 
             @Override
             protected void onFailure(Throwable exception) {
-                importProject(callback);
+                if (exception instanceof ServerException && ((ServerException)exception).getHTTPStatus() == 404) {
+                    importProject(callback);
+                } else {
+                   //TODO Show error message
```
sleshchenko: Can you show error message?

skabashnyuk:
LGTM

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
```diff
@@ -339,6 +352,14 @@ private HttpJsonHelper() {
      */
     public static class HttpJsonHelperImpl {
 
+        public void request(String url,
```
garagatyi: Looks like indentation is incorrect

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
```diff
@@ -132,6 +137,14 @@ public static String requestString(int timeout,
         return httpJsonHelperImpl.requestString(timeout, url, method, body, parameters);
     }
 
+    public static void request(String url,
```
garagatyi: Why do we need this extra method?

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
```diff
@@ -55,6 +55,11 @@
 
 
     //==============================================================
+    public static void request(Link link, Object body, Pair<String, ?>... parameters)
```
garagatyi: Why do we need that extra method

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
akorneta: it's ok

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
akorneta: it will be removed

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonHelper.java
akorneta: to delegate the methods of the private instance HttpJsonHelperImpl

garagatyi:
LGTM

PR #393, opened 2015-11-09
IDEX-3193: Activate go into featura for project nodes

AndrienkoAleksandr:


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/GoIntoAction.java
```diff
@@ -59,6 +59,6 @@ public void actionPerformed(ActionEvent e) {
     }
 
     private boolean isNodeSupportGoInto(Object node) {
-        return node != null && node instanceof Node && ((Node)node).getParent() != null && ((Node)node).supportGoInto();
+        return node != null && node instanceof Node && ((Node)node).supportGoInto();
```
vzhukovskii: checking for nullable parent was added specially, cause going into project node simply has no meaning

vzhukovskii: @vparfonov should we allow to go into project node?

AndrienkoAleksandr: But  it's has meaning when you works with few projects and you want to switch and work with one of them. By the way, Eclipse support go into for projects...

evoevodin: BTW `node != null` is useless check here

vzhukovskii: yes, you're right

AndrienkoAleksandr: node != null  and node.getParent != null this is different things...

AndrienkoAleksandr: Ok i agree with you about  node != null . I'll fix it

vparfonov:
ok

PR #394, opened 2015-11-10
Reset Go Into mode when new project created and when we try to scroll to file that is outside Go Into node

vzhukovskii:
@vparfonov little bug fix


vparfonov:
ok

PR #403, opened 2015-11-11
IDEX-3394: Make it possible to setup git branch and use correct sparse checkout

vzhukovskii:
@vparfonov @ashumilova


vparfonov:
ok

PR #404, opened 2015-11-11
IDEX-3330: Make vfs call to ext server when importing a project

vinokurig:
IDE making vfs calls to api but it have to make call to ext server
@vparfonov @skabashnyuk


skabashnyuk:
LGTM

vparfonov:
ok

PR #408, opened 2015-11-13
IDEX-3429: Remove links to GPL resources

evoevodin:
Remove `* along with this program; if not, see<http://www.gnu.org/licenses/>.\n" + "*/ \n` line from tests as it refers to GPL resources


evoevodin:
@skabashnyuk review please

skabashnyuk:
LGTM

PR #413, opened 2015-11-14
Expose direct API for converting DTOs to JSON objects

tareqhs:
this is more efficient than using the only API that is currently available
'toJson()'. It returns a string which means two redundant serializations must
be done if the JsonElement is needed directly:
DTO->JsonElement->String->JsonElement

Change-Id: I78cbfe0fd54b28d8b14ba53bc35b8cfd891f19bc


skabashnyuk:
For which purpose it can be used?

tareqhs:
Hi @skabashnyuk ,
in you code we interact with JSON objects in several places and might change them. doing this through toJson() will involve two additional redundant steps of conversion although the JSON objects are already provided directly beneath the APIs.

PR #415, opened 2015-11-16
IDEX-3511 : Remove deprecated class org.eclipse.che.commons.lang.Strings and its usage

rainf0x:
This is pull request issue https://jira.codenvycorp.com/browse/IDEX-3511

This pull request removes deprecated class  org.eclipse.che.commons.lang.Strings and use com.google.common.base.Strings com.google.common.base.Strings  instead of its.


che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
```diff
@@ -14,6 +14,7 @@
 import com.google.common.hash.HashFunction;
 import com.google.common.hash.Hashing;
 import com.google.common.io.ByteSource;
+import com.google.common.io.ByteStreams;
```
skabashnyuk: I think its better to remove FSMountPoint from this changes set.

skabashnyuk:
Other ok

che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
rainf0x: It's only import optimization.

evoevodin:
:+1:

PR #416, opened 2015-11-16
IDEX-3284: Show all Hotkey that available in Che

AndrienkoAleksandr:
Show all Hotkey that available in Che (including keys for Orion)


vparfonov:
ok

PR #418, opened 2015-11-17
IDEX-2981: Add '--recursive' parameter to git clone command

vinokurig:
We need git clone wirh --recursive to clone multi-module project with submodules
@vparfonov @skabashnyuk


platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectImporter.java
```diff
@@ -194,9 +198,12 @@ public void importSources(FolderEntry baseFolder, String location, Map<String, S
         }
     }
 
-    private void cloneRepository(GitConnection git, String remoteName, String url, DtoFactory dtoFactory)
+    private void cloneRepository(GitConnection git, String remoteName, boolean recursiveEnabled, String url, DtoFactory dtoFactory)
```
sleshchenko: Why do we need send instance of DtoFactory into method?
Please put recursiveEnabled parameter to end of parameters list.

vinokurig: fixed

PR #423, opened 2015-11-18
IDEX-2981: Add '--recursive' parameter to git clone command

vinokurig:
We need git clone wirh --recursive to clone multi-module project with submodules.
Will be merged only in 3.x
@vparfonov @skabashnyuk


vparfonov:
ok

skabashnyuk:
ok

PR #425, opened 2015-11-18
IDEX-3465: Fix opening images

AndrienkoAleksandr:
We can't open images, because inherited method getView() throws Unsupported Exception.


vparfonov:
ok

PR #430, opened 2015-11-19
Modules storing in project configuration

mshaposhnik:


platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/ModuleConfigImpl.java
```diff
@@ -114,6 +114,9 @@ public void setMixins(List<String> mixins) {
 
     @Override
     public List<ModuleConfig> getModules() {
+        if (modules == null) {
+            modules = new ArrayList<>();
+        }
         return modules;
```
skabashnyuk: firstNotNull

skabashnyuk: forget

skabashnyuk:
LGTM

PR #437, opened 2015-11-20
Fix import project

sleshchenko:
@vparfonov WDYT?

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/projectimport/wizard/ImportProjectNotificationSubscriberImpl.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/projectimport/wizard/ImportProjectNotificationSubscriberImpl.java
@@ -15,7 +15,8 @@ import com.google.gwt.json.client.JSONParser;
 import com.google.inject.Inject;
 import com.google.web.bindery.event.shared.EventBus;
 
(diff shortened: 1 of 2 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/projectimport/wizard/ImportProjectNotificationSubscriberImpl.java
```diff
@@ -51,12 +50,15 @@ public ImportProjectNotificationSubscriberImpl(CoreLocalizationConstant locale,
         this.notificationManager = notificationManager;
         this.workspaceId = appContext.getWorkspace().getId();
 
-        this.messageBus = messageBusProvider.getMessageBus();
+        messageBus = messageBusProvider.getMachineMessageBus();
```
vparfonov: Check this line please, i think it not need, it because we init messgeBus only after starting ExtServer

sleshchenko: It doesn't work without this line. I'll investigate cause of fail

sleshchenko: ImportProjectNotificationSubscriberImpl is not singleton and instance of this class can be constructed when extension server already is started.

PR #438, opened 2015-11-23
IDEX-3452: Remove public/private visibility switcher from project creation/importing wizard

azatsarynnyy:
@vparfonov


vparfonov:
ok

PR #445, opened 2015-11-25
IDEX-3011. Add ability show log from extension server.

RomanNikitenko:
Add getFileContent() method to MachineServiceClient


vparfonov:
ok

PR #447, opened 2015-11-25
IDEX-3396: Add ability to save workspace's snapshot

evoevodin:
This PR contains server/client logic for adding ability to create and recover from snapshots.

* Client
  * Added _Create Snapshot_ menu item to _Machine_ menu
  * Added start workspace logic which asks user if he wants to recover workspace from snapshot(if snapshot for this workspace exists) .
* Server
  * Changed `Snapshot` model (basically added _machineName_ and _envName_)
  * Added local implementation for `SnapshotDao` 
  * Added new API methods related to snapshots management and workspace recovering

@skabashnyuk, @garagatyi, @akorneta, @sleshchenko  please take a look

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
@@ -1,0 +1,93 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2015 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.ide.actions;
+
+import org.eclipse.che.api.promises.client.Operation;
+import org.eclipse.che.api.promises.client.OperationException;
+import org.eclipse.che.api.promises.client.PromiseError;
+import org.eclipse.che.api.workspace.gwt.client.WorkspaceServiceClient;
+import org.eclipse.che.ide.CoreLocalizationConstant;
+import org.eclipse.che.ide.api.notification.Notification;
+import org.eclipse.che.ide.api.notification.NotificationManager;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * Creates snapshot of the workspace using {@link WorkspaceServiceClient}.
+ *
+ * <p>This component is for managing notifications which are related to creating snapshot process.
+ *
+ * @author Yevhenii Voevodin
+ */
+@Singleton
+public class WorkspaceSnapshotCreator {
+
+    private final WorkspaceServiceClient   workspaceService;
+    private final NotificationManager      notificationManager;
+    private final CoreLocalizationConstant locale;
+
+    Notification notification;
+
+    @Inject
+    public WorkspaceSnapshotCreator(WorkspaceServiceClient workspaceService,
+                                    NotificationManager notificationManager,
+                                    CoreLocalizationConstant locale) {
+        this.workspaceService = workspaceService;
+        this.notificationManager = notificationManager;
+        this.locale = locale;
+    }
+
+    /**
+     * Changes notification state to finished with an error.
+     */
+    public void creationError(String message) {
+        notification.setMessage(message);
+        notification.setType(Notification.Type.ERROR);
+        notification.setStatus(Notification.Status.FINISHED);
+        notification = null;
+    }
+
+    /**
+     * Changes notification state to successfully finished.
+     */
+    public void successfullyCreated() {
+        notification.setMessage(locale.createSnapshotSuccess());
+        notification.setType(Notification.Type.INFO);
+        notification.setStatus(Notification.Status.FINISHED);
+        notification = null;
+    }
+
+    /**
+     * Returns true if workspace creation process is not done, otherwise when it is done - returns false
+     */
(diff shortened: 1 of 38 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
```diff
@@ -0,0 +1,88 @@
...
+        this.notificationManager = notificationManager;
+    }
+
+    /**
+     * Changes notification state to finished with an error.
+     */
+    public void creationError(String message) {
+        notification.setMessage(message);
+        notification.setType(Notification.Type.ERROR);
+        notification.setStatus(Notification.Status.FINISHED);
```
sleshchenko: You don't need set FINISHED status here.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
```diff
@@ -0,0 +1,88 @@
...
+        notification = null;
+    }
+
+    /**
+     * Changes notification state to successfully finished.
+     */
+    public void successfullyCreated() {
+        notification.setMessage("Snapshot successfully created");
+        notification.setType(Notification.Type.INFO);
+        notification.setStatus(Notification.Status.FINISHED);
```
sleshchenko: The same to https://github.com/codenvy/che-core/pull/447/files#r45862727

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Creates snapshot from workspace with given id and shows appropriate notification.
+     *
+     * @param workspaceId
+     *         id of the workspace to create snapshot from.
+     */
+    public void createSnapshot(String workspaceId) {
+        notification = new Notification("Creating snapshot...", Notification.Status.PROGRESS);
```
sleshchenko: you should move "Creating snapshot..." to localization

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
evoevodin: Why not? It looks logical for me. Do you have an example ?

azatsarynnyy: In fact, if you set type=ERROR then status will be set as FINISHED automatically.
But it's not obvious and I think it's better to set status as FINISHED explicitly.

sleshchenko: https://github.com/codenvy/che-core/blob/openshift/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/notification/Notification.java#L547-549

sleshchenko: We should have one way for this. This is first place where I see setting FINISHED status with setting ERROR type

azatsarynnyy: Really? ))
I can show you many such places in code

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
evoevodin: Yea you are right

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
sleshchenko: I think It would be better to have status FAILED. And instead of setting ERROR Type and FINISHED Status we should set only FAILED status. It will be clear and comfortably. But It is another story :)

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/bootstrap/WorkspaceComponent.java
```diff
@@ -323,6 +314,63 @@ public void apply(PromiseError arg) throws OperationException {
...
+     * <p> When "Ok" button is pressed - {@link WorkspaceServiceClient#recoverWorkspace(String, String, String) recovers workspace}
+     * <br>When "Cancel" button is pressed - {@link WorkspaceServiceClient#startById(String, String) starts workspace}
+     */
+    private void showRecoverWorkspaceConfirmDialog(final UsersWorkspaceDto workspace) {
+        dialogFactory.createConfirmDialog("Workspace recovering",
+                                          "Do you want to recover the workspace from snapshot?",
+                                          new ConfirmCallback() {
+                                              @Override
+                                              public void accepted() {
+                                                  handleWsStart(workspaceServiceClient.recoverWorkspace(workspace.getId(),
```
akorneta: You should handle an error during recovering workspace

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/WorkspaceSnapshotCreator.java
azatsarynnyy: I think it would be better if we would have some shortcut-method in *Notification* class which can set notification as finished in error state.

skabashnyuk:
Server side ok.
What about @vparfonov ?

PR #453, opened 2015-11-26
IDEX-3657: isInsideWorkTree should return false if directory is outside repo

dmytro-ndp:
@sleshchenko: please review changes.

### DIFF
--- a/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
+++ b/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
@@ -92,13 +92,14 @@ import java.util.regex.Pattern;
  */
 public class NativeGitConnection implements GitConnection {
 
-    private static final Pattern authErrorPattern =
-            Pattern.compile(
-                    ".*fatal: could not read (Username|Password) for '.*': No such device or address.*|" +
-                    ".*fatal: could not read (Username|Password) for '.*': Input/output error.*|" +
-                    ".*fatal: Authentication failed for '.*'.*|.*fatal: Could not read from remote repository\\.\\n\\nPlease make sure " +
-                    "you have the correct access rights\\nand the repository exists\\.\\n.*",
-                    Pattern.MULTILINE);
+    private static final Pattern authErrorPattern         =
+        Pattern.compile(
+            ".*fatal: could not read (Username|Password) for '.*': No such device or address.*|" +
+            ".*fatal: could not read (Username|Password) for '.*': Input/output error.*|" +
+            ".*fatal: Authentication failed for '.*'.*|.*fatal: Could not read from remote repository\\.\\n\\nPlease make sure " +
+            "you have the correct access rights\\nand the repository exists\\.\\n.*",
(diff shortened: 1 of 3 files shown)


che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
```diff
@@ -323,12 +323,20 @@ public boolean isInsideWorkTree() throws GitException {
...
-        return Boolean.valueOf(output);
+        try {
+            emptyGitCommand.setNextParameter("rev-parse")
+                           .setNextParameter("--is-inside-work-tree")
+                           .execute();
+                   
+            final String output = emptyGitCommand.getText();
+            return Boolean.valueOf(output);
+        } catch(GitException ge) {
+            if (ge.getMessage().matches("(?s).*fatal: Not a git repository.*(?s).*")) {
```
garagatyi: It would be better to compile it as static field

dmytro-ndp:
@garagatyi: thank you for remark! I have fixed that.

che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
```diff
@@ -323,12 +324,21 @@ public boolean isInsideWorkTree() throws GitException {
...
+        try {
+            emptyGitCommand.setNextParameter("rev-parse")
+                           .setNextParameter("--is-inside-work-tree")
+                           .execute();
+                   
+            final String output = emptyGitCommand.getText();
+            return Boolean.valueOf(output);
+        } catch(GitException ge) {
+            String msg = ge.getMessage();
+            if (msg != null && notInGitRepoErrorPattern.matcher(msg).find()) {
```
garagatyi: I think method `match()` fits better in that case

dmytro-ndp: why?

garagatyi: I meant method `matches()`
Here is part of javadocs of these methods:
 - find() `Attempts to find the next subsequence of the input sequence that matches the pattern.`
 - matches() `Attempts to match the entire region against the pattern.` 
And in this case we check if entire string matches.

sleshchenko: +1 to use match is this case

sleshchenko:
OK for me

che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
dmytro-ndp: matches() doesn't work in this case as expected :-(

dmytro-ndp:
@garagatyi: I have replaced matcher() on find()

garagatyi:
OK for me

skabashnyuk:
ok

PR #457, opened 2015-11-27
IDEX-3468: Replace ProjectDescriptor in according to new project model

dimasnurenko:
@vparfonov @evidolob @azatsarynnyy @sleshchenko  @garagatyi @skabashnyuk @evoevodin please review pull request


sleshchenko:
Please adopt https://github.com/codenvy/plugin-openshift to this changes

platform-api/che-core-api-infrastructure-local/src/main/java/org/eclipse/che/api/local/ProjectConfigAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public JsonElement serialize(ProjectConfig moduleConfig, Type type, JsonSerializationContext context) {
+        return serializeModule(moduleConfig, context);
+    }
+
+    private JsonElement serializeModule(ProjectConfig moduleConfig, JsonSerializationContext context) {
+        JsonObject object = new JsonObject();
+
```
vparfonov: Need to check that we need store to the Workspace. Look like not all attributes must be stored because some of them are calculate runtime.

platform-api/che-core-api-infrastructure-local/src/main/java/org/eclipse/che/api/local/ProjectConfigAdapter.java
```diff
@@ -0,0 +1,67 @@
...
+        object.addProperty("name", moduleConfig.getName());
+        object.addProperty("path", moduleConfig.getPath());
+        object.addProperty("type", moduleConfig.getType());
+        object.addProperty("content_root", moduleConfig.getContentRoot());
+        object.add("modules", serializeModules(moduleConfig, context));
+        object.add("source", context.serialize(moduleConfig.getSource()));
```
vparfonov: add  object.add("attributes", context.serialize(moduleConfig.getAttributes()));

vparfonov:
try use interface ProjectConfig wherever possible

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/project/shared/ContentRootHelper.java
```diff
@@ -43,21 +41,21 @@
      * in Eclipse IDE.
      *
      * @param node
-     *         node to process, should be {@link ProjectDescriptorNode}
+     *         node to process, should be {@link org.eclipse.che.ide.project.node.ProjectNode}
```
evoevodin: Replace fqn `{@link org.eclipse.che.ide.project.node.ProjectNode}` with `{@link ProjectNode}`

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/project/type/ProjectType.java
```diff
@@ -25,7 +25,7 @@
 
     List<? extends Attribute> getAttributes();
 
-    List<? extends ProjectType> getParents();
+    List<ProjectType> getParents();
```
evoevodin: Are you sure about this change? It seems that we don't have _DTO_ implemenation of `ProjecType`, but if we do it won't work in this way.

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/project/type/ProjectType.java
```diff
@@ -33,7 +33,7 @@
 
     Attribute getAttribute(String name);
 
-    boolean getCanBeMixin();
+    boolean canBeMixin();
```
evoevodin: Dto doesn't support `can` like methods, `getCanBeMixin` is also not so good as it should be, so either find another name for this method or implement `can` - like methods in _DTO_ framework :smile:

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/project/type/ProjectType.java
dimasnurenko: We don't have DTO implementation of ProjectType

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
```diff
@@ -439,134 +430,127 @@ public ProjectConfig getProjectConfig(Project project) throws ServerException, P
...
-                        throw new ProjectTypeConstraintException("Required attribute value is initialized with null value " + attr.getId());
-                    }
-                } else {
-                    // add constants
-                    projectConfig.getAttributes().put(attr.getName(), attr.getValue().getList());
-                }
-            }
-        }
+        ArrayList<String> ms = new ArrayList<>();
+        ms.addAll(types.getMixins().keySet());
```
evoevodin: You can do it in a single line `new ArrayList<>(types.getMixins().keySet())` which also improves memory usage.

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/project/type/ProjectType.java
dimasnurenko: ProjectType is just interface on server side and it doesn't have DTO implementation.

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
```diff
@@ -578,26 +562,51 @@ public void updateProjectConfig(Project project, ProjectConfig config)
...
 
-    private void getModulesRecursive(ModuleConfig moduleConfig, List<ModuleConfigDto> projectModules) {
-        List<ModuleConfigDto> modules = new ArrayList<>();
+    private void filterAttributesRecursive(ProjectConfigDto projectConfig,
+                                           ProjectTypes projectTypes,
+                                           FolderEntry folderEntry) throws ValueStorageException,
+                                                                           ProjectTypeConstraintException,
+                                                                           ServerException,
+                                                                           ForbiddenException {
+        Map<String, List<String>> runtimeAttributes = new HashMap<>();
```
evoevodin: Aa i can see `runtimeAttributes` are used only in line *580*, so if you don't need this attributes you should change the logic of `filterAttributes` or use `Collection.emptyMap()` which may also require `filterAttributes` method changes as it is immutable.

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
```diff
@@ -936,82 +945,104 @@ public VirtualFileEntry rename(String workspace, String path, String newName, St
...
+    }
 
-            eventService.publish(new ProjectItemModifiedEvent(ProjectItemModifiedEvent.EventType.DELETED, workspace,
-                                                              projectPath(entry.getPath()), entry.getPath(), entry.isFolder()));
-            entry.remove();
+    private void deleteProjectFromWorkspace(ProjectConfig project, String workspaceId) throws ServerException,
+                                                                                              ForbiddenException {
+        String projectName = project.getName();
+
+        sendRecuestToDeleteProject(workspaceId, projectName);
```
evoevodin: i don't like this method name even without typo. Think more about it.

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
```diff
@@ -1019,8 +1050,8 @@ private void deleteProjectFromWorkspace(String wsId, String projectName) throws
 
         try {
             HttpJsonHelper.request(null, link);
-        } catch (IOException | ApiException e) {
-            throw new ServerException(e);
+        } catch (IOException | ApiException exception) {
```
evoevodin: Just interesting: what is wrong with `e` ? 
In this case i prefer to use either `ex` or separated `ioEx` and `apiEx` names, but i don't think that `expection` is better than `e`

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
```diff
@@ -1019,8 +1050,8 @@ private void deleteProjectFromWorkspace(String wsId, String projectName) throws
...
+            throw new ServerException(exception);
```
evoevodin: In this case it is better to throw `new ServerException(expection.getLocalizedMessage(), exception)`

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DtoConverter.java
```diff
@@ -158,170 +126,109 @@ public static ItemReference toItemReference(FolderEntry folder,
...
+    /**
+     * The method tries to provide as much as possible information about project.If get error then save information about error
+     * with 'problems' field in ProjectConfigDto.
+     *
+     * @param project
+     *         project from which we need get information
+     * @param serviceUriBuilder
+     *         service for building URI
+     * @return an instance of {@link ProjectConfigDto}
+     * @throws InvalidValueException
```
evoevodin: Describe exception javadoc as well. [Doc](https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#javadoc) says that _@param, @return, @throws or @deprecated must not be appeared without of description_

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DtoConverter.java
```diff
@@ -158,170 +126,109 @@ public static ItemReference toItemReference(FolderEntry folder,
...
+     */
+    public static ProjectConfigDto toProjectConfig(Project project,
+                                                   UriBuilder serviceUriBuilder) throws ForbiddenException,
+                                                                                        ServerException,
+                                                                                        ProjectTypeConstraintException,
+                                                                                        ValueStorageException {
+        String name = project.getName();
+        String path = project.getPath();
+
+        ProjectConfigDto projectConfigDto = DtoFactory.getInstance().createDto(ProjectConfigDto.class);
```
evoevodin: Just a hint: In this case `DtoFactory.newDto(ProjectConfigDto.class)` looks much better

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
dimasnurenko: It is just action in opposition to assign "one letter name" to variables. And i think in this case name exception is better then 'e'.

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DtoConverter.java
```diff
@@ -158,170 +126,109 @@ public static ItemReference toItemReference(FolderEntry folder,
...
+            projectConfigDto.withContentRoot(config.getContentRoot());
+
+            projectConfigDto.withType(config.getType());
+            projectConfigDto.withSource(toSourceDto(config.getSource()));
+            projectConfigDto.setContentRoot(project.getContentRoot());
+        } catch (ServerException | ValueStorageException | ProjectTypeConstraintException exception) {
+            projectConfigDto.withType(BaseProjectType.ID).withType("blank");
+            ProjectProblem projectProblem = DtoFactory.getInstance()
+                                                      .createDto(ProjectProblem.class)
+                                                      .withCode(1)
```
evoevodin: `1` what is 1?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
```diff
@@ -256,14 +266,27 @@ VirtualFileEntry rename(String workspace, String path, String newName, String ne
...
+     * Deletes module from project and updates project in workspace.
+     *
+     * @param workspaceId
+     *         workspace id to get project from workspace
+     * @param pathToParent
+     *         path to parent of deleted module
+     * @param pathToModule
+     *         path to module which will be deleted
+     */
+    void deleteModule(String workspaceId, String pathToParent, String pathToModule) throws ServerException,
```
evoevodin: What about exceptions in javadoc?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -610,13 +603,25 @@ public Response updateFile(@ApiParam(value = "Workspace ID", required = true)
...
 
-        if (!projectManager.delete(workspace, path)) {
-            throw new NotFoundException(String.format("Path '%s' doesn't exist.", path));
-        }
+        projectManager.delete(workspace, path);
+    }
+
+    @DELETE
+    @Path("module/{path:.*}")
+    public void deleteModule(@ApiParam(value = "Workspace ID", required = true)
```
evoevodin: `@PathParam` doesn't have to be required, it is required by default, remove `required = true` + use `@ApiParam("Workspace ID")` as `value` will be excessive too. The same for the other `@PathParam`s

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -610,13 +603,25 @@ public Response updateFile(@ApiParam(value = "Workspace ID", required = true)
...
         filesBuffer.addToBufferRecursive(getVirtualFileEntry(workspace, path).getVirtualFile());
 
-        if (!projectManager.delete(workspace, path)) {
-            throw new NotFoundException(String.format("Path '%s' doesn't exist.", path));
-        }
+        projectManager.delete(workspace, path);
+    }
+
+    @DELETE
+    @Path("module/{path:.*}")
```
evoevodin: If you are defining parameters with swagger annotations, please don't forget to define `@ApiOperation` and `@ApiReponses`

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DtoConverter.java
dimasnurenko: In this case i can't explain you why we use '1'. I don't write this code, i just ported it and decided don't change it, and i couldn't find any java doc which would explain why we use '1', as project problem code.

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/type/AbstractAttribute.java
```diff
@@ -33,7 +33,7 @@ protected Attribute(String projectType, String name, String description, boolean
 
 
     public String getId() {
-        return projectType+":"+name;
+        return projectType + ":" + name;
```
evoevodin: Just a hint: use character literals for concatenations like this as it uses `StringBuilder` in background and character append worth less than string append.

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/ProjectConfigImpl.java
```diff
@@ -47,36 +156,83 @@ public SourceStorage getSource() {
...
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (!(o instanceof ProjectConfigImpl)) return false;
-        if (!super.equals(o)) return false;
         final ProjectConfigImpl other = (ProjectConfigImpl)o;
-        return Objects.equals(storage, other.storage);
+        return Objects.equals(name, other.name) &&
```
evoevodin: '&&' must go in a new line see [doc](https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#for-any-other-operator-the-break-comes-before-the-symbol)

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/shared/dto/util/ProjectConfigUtil.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Finds modules in project recursively.
+     *
+     * @param config
+     *         project in which we need to find a module
+     * @param pathToModule
+     *         path to module which we need to find
+     * @return an instance {@link ProjectConfigDto} which describes found module or <code>null</code> if module haven't found
+     */
+    public static ProjectConfigDto findModuleRecursive(ProjectConfigDto config, String pathToModule) {
```
evoevodin: I don't think that `Recursive` is a part of the method name.

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/shared/dto/util/ProjectConfigUtil.java
```diff
@@ -0,0 +1,50 @@
...
+            }
+
+            ProjectConfigDto foundConfig = findModuleRecursive(configDto, pathToModule);
+
+            if (foundConfig != null) {
+                return foundConfig;
+            }
+        }
+
+        return null;
```
evoevodin: It is better to use `Optional` in the case like this

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/shared/dto/util/ProjectConfigUtil.java
```diff
@@ -0,0 +1,50 @@
...
+package org.eclipse.che.api.workspace.shared.dto.util;
+
+import org.eclipse.che.api.workspace.shared.dto.ProjectConfigDto;
+
+/**
+ * Special util class which is necessary to generate methods in {@link ProjectConfigDto}
+ *
+ * @author Dmitry Shnurenko
+ */
+public class ProjectConfigUtil {
```
evoevodin: Define util classes as `final` + add default private constructor.

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/shared/dto/util/ProjectConfigUtil.java
evoevodin: What if config is `null` ?

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/shared/dto/util/ProjectConfigUtil.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Finds modules in project recursively.
+     *
+     * @param config
+     *         project in which we need to find a module
+     * @param pathToModule
+     *         path to module which we need to find
+     * @return an instance {@link ProjectConfigDto} which describes found module or <code>null</code> if module haven't found
```
evoevodin: If you want to use _code-like_ font for the `null`, use `{@code null}` instead

evoevodin:
As far as i understand this PR is about refactoring, next time please add some comment which explains your changes.

I added my remarks, everything else looks ok for me.

PR #459, opened 2015-11-29
General injection-based filtering of system files in VFS

tareqhs:
Currently, system-level filtering of virtual files is done through hard-coded checks in VFS projects. This is bad because the actual source of truth for such checks is in layers that are higher than VFS, e.g. hiding misc.xml from che-core-api-project.

A better solution is to allow defining decoupled injection-based filters that allow filtering VFS paths from their correct sources of truth in components at higher layers.


vparfonov:
ok

tareqhs:
hi @vparfonov ,
thank you. can this be merged?

skabashnyuk:
ok

rainf0x:
Eclipse CLA confirmed for tareq.sha@gmail.com Eclipse CLA confirmed

PR #465, opened 2015-12-01
IDEX-3713: GitMixin is not propagated to modules.

vparfonov:
Problems is not correct deternination is module under git repository or not.
Now we detect it in two steps:
- firts we run git rev-parse --is-inside-work-tree for detecting is any git repository exist
- second we check that this repo under our virtaul file system control

### DIFF
--- a/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
+++ b/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
@@ -11,6 +11,8 @@
 package org.eclipse.che.git.impl.nativegit;
 
 
+import com.google.common.base.Strings;
+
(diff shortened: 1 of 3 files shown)


vparfonov:
@skabashnyuk @sleshchenko

che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
```diff
@@ -112,9 +116,9 @@
      * @throws GitException
      *         when some error occurs
      */
-    public NativeGitConnection(File repository, GitSshScriptProvider gitSshScriptProvider,
+    public NativeGitConnection(File mountRoot, File repository, GitSshScriptProvider gitSshScriptProvider,
```
skabashnyuk: can you describe what is mounRoot and what is repository?

sleshchenko:
Please add test for this code

PR #475, opened 2015-12-03
Remove deprecated annotation of update project method

sleshchenko:
@vparfonov This method is not deprecated for now


vparfonov:
ok

PR #476, opened 2015-12-03
Remove gwt compiling for IE9

sleshchenko:
@vparfonov @evidolob @azatsarynnyy I removed settings for ie9. 
And I have question: Will it work https://github.com/codenvy/che-core/blob/784fed052d53ea3d98c1232936afeb68a2d0e90b/commons/che-core-commons-gwt/src/main/resources/org/eclipse/che/ide/Commons.gwt.xml#L46-L54 if we don't set user agent to android and iphone?

### DIFF
--- a/commons/che-core-commons-gwt/src/main/resources/org/eclipse/che/ide/Commons.gwt.xml
+++ b/commons/che-core-commons-gwt/src/main/resources/org/eclipse/che/ide/Commons.gwt.xml
@@ -19,33 +19,22 @@
 
     <source path=""/>
 
+    <set-property name="user.agent" value="safari, gecko1_8"/>
+
     <!-- Allow for conditional compilation per user.agent with a single "if": -->
-    <set-property name="user.agent" value="safari, gecko1_8, ie9"/>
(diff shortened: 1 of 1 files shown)


vparfonov:
Ok for merging 
Sorry, but I don't know

evidolob:
What do you mean: "Will it work" ?

sleshchenko:
I mean "Will it compiled for android and for iphone?"

evidolob:
No it mean that if user agent is iphone use IPhoneImpl, not a compile new permutation

sleshchenko:
So We can just remove it?

evidolob:
if we not interesting in detecting iphone, yes

sleshchenko:
iphone can be detected even if we do not add user.agent="iphone" https://github.com/codenvy/che-core/pull/476/files#diff-4f669b01dc4a4299e39b83ef32a6380dR22?

evidolob:
they use 'mobile.user.agent' not a 'user.agent', so if you test it, I think you can delete them

sleshchenko:
you're right. I'll investigate this.

PR #477, opened 2015-12-03
set alt+cmd+shift+n hotkey for NavigateToFileAction for Mac

evidolob:


vparfonov:
ok

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/core/StandardComponentInitializer.java
```diff
@@ -497,7 +498,11 @@ public void initialize() {
         loaderToolbarGroup.add(loaderAction);
 
         // Define hot-keys
-        keyBinding.getGlobal().addKey(new KeyBuilder().action().alt().charCode('n').build(), "navigateToFile");
+        if(UserAgent.isMac()){
+            keyBinding.getGlobal().addKey(new KeyBuilder().action().alt().charCode('N').build(), "navigateToFile");
```
sleshchenko: Will be it work for Linux?

evidolob: Now it's working?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/core/StandardComponentInitializer.java
```diff
@@ -497,7 +498,11 @@ public void initialize() {
         loaderToolbarGroup.add(loaderAction);
 
         // Define hot-keys
-        keyBinding.getGlobal().addKey(new KeyBuilder().action().alt().charCode('n').build(), "navigateToFile");
+        if(UserAgent.isMac()){
```
sleshchenko: BTY I think it would be better to write
```
keyBinding.getGlobal().addKey(new KeyBuilder().action().alt().charCode(UserAgent.isMac()? 'N' : 'n').build(), "navigateToFile");
```

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/core/StandardComponentInitializer.java
sleshchenko: Sorry. What do you mean?

evidolob: I not change anything related to Linux or Windows

sleshchenko: But If 'N' will be work for Linux and Windows and Mac why do we need redundant checking for MacOS

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/core/StandardComponentInitializer.java
evidolob: I don't think so,  for me it's more difficult to understand that for mac uses differed hotkey

azatsarynnyy: In this case it would be better to write something like this:

int charCode = UserAgent.isMac() ? 'N' : 'n';
keyBinding.getGlobal().addKey(new KeyBuilder().action().alt().charCode(charCode).build(), "navigateToFile");

I think it will be more readable and understandable.
WDYT, guys?

sleshchenko: if we name it something like navigateToFileCharCode it will be fine I think

PR #480, opened 2015-12-03
IDEX-3574 make the case of zeroClipboard the same like we have in master

skabashnyuk:
This will allow to checkout to branch 4.0 from master in che in case insensitive filesystems


evoevodin:
(doc)[https://github.com/codenvy/codenvy-dev-resources/wiki/Naming#package-names]

benoitf:
yes we shouldn't have to switch back to some uppercase in the package

benoitf:
you may also get directly the correct branch by using git clone -b <branchName> <repo-url>
like git clone -b 4.0 https://github.com/codenvy/che che in the terminal of Che

skabashnyuk:
I  can do that. But it would be hard to explain to other users
@vparfonov @gazarenkov  voted to rollback this changes in 4.0 not in master.

benoitf:
can we switch master to 4.0 ? it looks like strange to put back wrong names (camelCase for the package name)
Or then packages need to be fixed in 3.x as well

PR #484, opened 2015-12-04
CLDIDE-2639: move performAction to ActionManager, adopt OpenFileAction

akorneta:


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/ActionManagerImpl.java
```diff
@@ -39,18 +43,25 @@
...
+/**
+ * @author Evgen Vidolob
+ */
 public class ActionManagerImpl implements ActionManager {
+    public static final String[] EMPTY_ARRAY = new String[0];
+
+    private final Map<String, Object>      myId2Action = new HashMap<>();
+    private final Map<String, Set<String>> myPlugin2Id = new HashMap<>();
+    private final Map<String, Integer>     myId2Index  = new HashMap<>();
+    private final Map<Object, String>      myAction2Id = new HashMap<>();
```
evoevodin: Align these fields by initializing them in the constructor

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/action/ActionManager.java
```diff
@@ -114,4 +115,16 @@
...
+     * Performs registered action by given id.
+     *
+     * <p>Note that if action is not registered nothing will be done
+     *
+     * @param actionId
+     *         the id of action that will be performed
+     * @param parameters
+     *         the parameters with which the action is invoked or updated
+     */
+    void performAction(String actionId, Map<String, String> parameters);
```
evoevodin: What if `actionId` or `parameters` map is null ? How should an implementation behave in this case.
It is always nice to mark parameters with `@Nullable` and `@NotNull` annotations + define behaviour.

mshaposhnik:
Ok

skabashnyuk:
ok

PR #485, opened 2015-12-06
Null "generatorDescription" safety in ProjectService.createModule()

tareqhs:
If "generatorDescription" is null when creating a module via ProjectService, it fails with 500 NullPointerException. This patch adds null safety.

### DIFF
--- a/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
+++ b/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
@@ -356,10 +356,17 @@ public class ProjectService extends Service {
                                           NewProject newProject)
             throws NotFoundException, ConflictException, ForbiddenException, ServerException {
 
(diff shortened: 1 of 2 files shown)


skabashnyuk:
What was removed in test?

tareqhs:
Hi @skabashnyuk 
The existing test contains a commented-out code that specifies an explicit generator to use, before my change the generator description is empty - i removed it so it is equivalent now. without my change a missing generatorDescription causes NPE. Do you think a dedicated test for this is required? I can add it.
thank you

skabashnyuk:
Before we have test with withGeneratorDescription(generatorDescription); you change some part related to generatorDescription . So yes I do think we need both tests.

tareqhs:
@skabashnyuk Done

vparfonov:
ok

skabashnyuk:
ok

rainf0x:
Eclipse CLA confirmed for tareq.sha@gmail.com Eclipse CLA confirmed

PR #486, opened 2015-12-07
IDEX-3747: Improve attribute filter. Fix update project with openshift mixin

dimasnurenko:
@vparfonov @sleshchenko


platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/AttributeFilter.java
```diff
@@ -0,0 +1,196 @@
...
+     */
+    public void addRuntimeAttributesToProject(ProjectConfigDto projectConfig,
+                                              FolderEntry projectFolder) throws ProjectTypeConstraintException,
+                                                                                ForbiddenException,
+                                                                                ValueStorageException,
+                                                                                ServerException {
+        addAttributesToProject(projectConfig, projectFolder, RUNTIME);
+    }
+
+    enum AttributeType {
```
sleshchenko: Do you need package private modifier for this enum?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/AttributeFilter.java
```diff
@@ -0,0 +1,196 @@
...
+     * @throws ServerException
+     *         error occurs during we get project from workspace
+     * @throws ForbiddenException
+     *         if user which perform operation doesn't have access to the requested project
+     * @throws ProjectTypeConstraintException
+     *         if we create project types without required parameters
+     * @throws ValueStorageException
+     *         when we have exception during getting values from value factory in attribute
+     */
+    public void addRuntimeAttributesToProject(ProjectConfigDto projectConfig,
```
sleshchenko: Please place public method above to of private methods

vparfonov:
ok

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/AttributeFilter.java
dimasnurenko: Yes. I use this enum only in this class for convenience and to do code more readable.

sleshchenko: If you use only for this class It would be better to use private. No?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/AttributeFilter.java
dimasnurenko: i think private methods have to be located as near as possible to places where they are used

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/AttributeFilter.java
dimasnurenko: Yes. You are right.

PR #489, opened 2015-12-08
IDEX-3650: Fix behavior of ExtServerStateController

akorneta:


skabashnyuk:
ok. @vparfonov ?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/projectimport/wizard/ImportProjectNotificationSubscriberImpl.java
```diff
@@ -13,117 +13,135 @@
...
+        messageBusPromise.then(new Operation<MessageBus>() {
+            @Override
+            public void apply(MessageBus messageBus) throws OperationException {
+                try {
+                    messageBus.unsubscribe(wsChannel, subscriptionHandler);
+                } catch (WebSocketException e) {
+                    Log.error(getClass(), e);
+                }
+                notification.setStatus(Notification.Status.FINISHED);
+                notification.setType(Notification.Type.ERROR);
```
evoevodin: I don't really know the truth but i think that in this case it would be better to set notification type before setting notification status, as `FINISHED` looks like trigger for notification progress stop.

vparfonov:
ok

PR #490, opened 2015-12-08
Support for maxItems, skipCount query parameters - Search API

Ori-Libhaber:
1. Added support for <b>maxItems</b> and <b>skipCount</b> query parameters, as described in email thread:<br>
<b>[che-dev] Search API</b>
2. Search result is now <b>encapsulated in SearchResult object</b> which provides additional metadata on search.


skabashnyuk:
You've closed it because of this https://github.com/codenvy/che-core/pull/501?

Ori-Libhaber:
Yes, that's the reason :)

PR #494, opened 2015-12-08
Change expand icon and hover state for tree node

RomanNikitenko:
Change expand icon and hover state for tree node


vparfonov:
ok

PR #499, opened 2015-12-09
Add User DAO getByName method

mshaposhnik:


platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/dao/UserDao.java
```diff
@@ -73,10 +77,21 @@
...
-     * @throws org.eclipse.che.api.core.NotFoundException
+     * @throws NotFoundException
      *         when user doesn't exist
      */
     User getById(String id) throws NotFoundException, ServerException;
+
+    /**
+     * Gets user from persistent layer by his username
+     *
+     * @param userName
```
sleshchenko: I think It would be better to rename this parameter to user if it can equals to name or id.

mshaposhnik: No, there can be only name (fixed javadoc)

skabashnyuk:
ok

PR #502, opened 2015-12-10
Source level

vzhukovskii:
It will avoid static analyzers in IDE's to propose use java 8 features. Because java 8 isn't available on the client side.
@vparfonov @skabashnyuk WDYT about it?


sleshchenko:
:+1: But merge these commits to one

sleshchenko:
P.S. I think we should set compiler and target level to 1.7 in all modules which contains only code of client side.

skabashnyuk:
ok for me. @vparfonov ?

vparfonov:
+1

PR #503, opened 2015-12-10
IDEX-3802: Configure project as blank by default after import

dimasnurenko:
@vparfonov @evoevodin @mshaposhnik @sleshchenko

### DIFF
--- a/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
+++ b/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
@@ -60,6 +60,7 @@ import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
+import javax.validation.constraints.NotNull;
 import javax.ws.rs.core.UriBuilder;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -554,7 +555,8 @@ public final class DefaultProjectManager implements ProjectManager {
         }
     }
 
-    private ProjectConfigDto getProjectFromWorkspace(String wsId, String projectPath) throws ServerException {
+    @Override
+    public ProjectConfigDto getProjectFromWorkspace(@NotNull String wsId, @NotNull String projectPath) throws ServerException {
         final UsersWorkspaceDto usersWorkspaceDto = getWorkspace(wsId);
         final String path = projectPath.startsWith("/") ? projectPath : "/" + projectPath;
         for (ProjectConfigDto projectConfig : usersWorkspaceDto.getProjects()) {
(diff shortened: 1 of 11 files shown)


vparfonov:
ok for me, @mshaposhnik @skabashnyuk  ?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
```diff
@@ -63,6 +64,29 @@
...
+     *         if other error occurs
+     * @throws IOException
+     * @throws ForbiddenException
+     *         if user which perform operation doesn't have access to the requested project
+     * @throws NotFoundException
+     *         when project not found
+     * @throws ConflictException
+     *         if operation causes conflict
+     */
+    void configureProjectAfterImport(SourceStorageDto storage, String workspaceId) throws ServerException,
```
evoevodin: I don't see the contract for `null` values, you should define this contract see [example](https://github.com/codenvy/che-core/blob/4.0/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonRequest.java#L119-L122)

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
```diff
@@ -63,6 +64,29 @@
...
     /**
+     * Configure project after import. If project wasn't found in workspace it will be configured as blank.
+     *
+     * @param storage
+     *         storage from which project was imported
+     * @param workspaceId
+     *         workspace id
+     * @throws ServerException
+     *         if other error occurs
+     * @throws IOException
```
evoevodin: Add [description](https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#generally) because `@param`, `@return`, `@throws` or `@deprecated` must not be appear without description

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
evoevodin: Also does it really matter to configure project exactly after the import ?
This method may appear in sort of *Trigger* interface, but it is a bit strange to see method with name like this in the `ProjectManager` interface.

evoevodin:
You added new API method, and didn't test it.
Can you please add tests?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/DefaultProjectManager.java
```diff
@@ -527,6 +528,42 @@ private ProjectConfigDto getProjectFromWorkspace(String wsId, String projectPath
...
+        if (projectFromWorkspace != null) {
+            return;
+        }
+
+        ProjectConfigDto blankProject = newDto(ProjectConfigDto.class).withName(projectName)
+                                                                      .withPath(projectPath)
+                                                                      .withType("blank")
+                                                                      .withContentRoot("")
+                                                                      .withDescription("")
+                                                                      .withAttributes(Collections.emptyMap())
```
evoevodin: You don't have to use empty collections for DTO objects.

vparfonov:
ok

mshaposhnik:
Looks good to me

PR #505, opened 2015-12-10
CLDIDE-2636: Improve account model

evoevodin:
In this PR i: 
* Improved account model
* Added `AccountDao.exists` method
* Fixed `WorkspaceServiceClient` behaviour, it was sending 'null' string instead of null object in query parameters.

### DIFF
--- a/platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountService.java
+++ b/platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountService.java
@@ -158,9 +158,7 @@ public class AccountService extends Service {
         } catch (NotFoundException ignored) {
         }
         final String accountId = NameGenerator.generate(Account.class.getSimpleName().toLowerCase(), Constants.ID_LENGTH);
-        final Account account = new Account().withId(accountId)
-                                             .withName(newAccount.getName())
-                                             .withAttributes(newAccount.getAttributes());
+        final Account account = new Account(accountId, newAccount.getName(), null, newAccount.getAttributes());
 
         accountDao.create(account);
         if (current != null) {
--- a/platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/dao/AccountDao.java
+++ b/platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/dao/AccountDao.java
@@ -15,7 +15,10 @@ import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.model.workspace.UsersWorkspace;
 
+import javax.validation.constraints.NotNull;
 import java.util.List;
+
+import static java.util.Objects.requireNonNull;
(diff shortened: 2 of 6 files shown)


skabashnyuk:
ok

akorneta:
ok

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountService.java
```diff
@@ -158,9 +158,7 @@ public Response create(@Context SecurityContext securityContext,
         } catch (NotFoundException ignored) {
         }
         final String accountId = NameGenerator.generate(Account.class.getSimpleName().toLowerCase(), Constants.ID_LENGTH);
-        final Account account = new Account().withId(accountId)
-                                             .withName(newAccount.getName())
-                                             .withAttributes(newAccount.getAttributes());
+        final Account account = new Account(accountId, newAccount.getName());
```
garagatyi: why you omit setting account attributes?

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/dao/AccountDao.java
```diff
@@ -151,4 +154,50 @@ default boolean isWorkspaceRegistered(String workspaceId) throws ServerException
...
+     *         when {@code accountId} is null
+     * @throws ServerException
+     *         when any other error occurs
+     * @see #getById(String)
+     */
+    default boolean exist(@NotNull String accountId) throws ServerException {
+        try {
+            getById(requireNonNull(accountId, "Required non-null account id"));
+            return true;
+        } catch (NotFoundException ignored) {
```
garagatyi: It is not ignored exception

platform-api/che-core-api-account/src/test/java/org/eclipse/che/api/account/AccountServiceTest.java
```diff
@@ -368,7 +352,7 @@ public void shouldBeAbleToGetAccountByName() throws Exception {
     }
 
     @DataProvider(name = "roleProvider")
-    public String[][] roleProvider() {
+    public Object[][] roleProvider() {
```
garagatyi: Can you explain why you changed that?

platform-api/che-core-api-account/src/main/java/org/eclipse/che/api/account/server/AccountService.java
evoevodin: Its missrefactoring, thank you

platform-api/che-core-api-account/src/test/java/org/eclipse/che/api/account/AccountServiceTest.java
evoevodin: `@DataProvider` javadoc says:  _The annotated method must return an `Object[][]` where each `Object[]` can be assigned the parameter list of the test method._

garagatyi:
ok

PR #507, opened 2015-12-10
Extract HTTP code of HttpJsonHelper to a statically injected interface

tareqhs:
This is an experimental change for discussion.

Here i suggest moving the core HTTP request/response code of HttpJsonHelper to a separate optional statically injected interface 'HttpHelper' The current behavior isn't affected as HttpJsonHelper defaults to using an implementation that contain the current code that is based on HttpURLConnection. Custom assemblies optionally request a static injection on HttpJsonHelper to provide a custom implementation and conveniently apply system-wide management of HTTP connectivity.

This should be temporary solution towards an incremental one that solves the issue of scattered and non-managed HTTP connectivity in Che (both JSON and non-JSON requests).

Relevant mailing list threads:
https://dev.eclipse.org/mhonarc/lists/che-dev/msg00930.html
https://dev.eclipse.org/mhonarc/lists/che-dev/msg00931.html

### DIFF
--- a/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpHelper.java
+++ b/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpHelper.java
@@ -1,0 +1,128 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2015 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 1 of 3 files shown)


skabashnyuk:
https://github.com/codenvy/che-core/blob/4.0/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpJsonRequest.java is this the same that you are proposing?

tareqhs:
The purpose is different, as discussed thoroughly in the provided mailing thread.
The purpose of Yevhenii's patch in https://github.com/codenvy/che-core/commit/2aa05905269a75d358ac282260d927bdca232d77 is to provide a 'chain builder' API for sending/receiving JSON requests as an alternative to HttpJsonHelper.
My patch addresses the core HTTP mechanism itself, which is also used for non-JSON requests. I would like to manage the requests and intercept them system-wide, or possibly even have a non-HttpURLConnection implementation.

PR #510, opened 2015-12-11
IDEX-3667: create new command wizard

olexii4:
@ashumilova  WDYT?


vzhukovskii:
Good!

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/theme/Theme.java
```diff
@@ -1306,4 +1306,18 @@
      */
     String loaderProgressStatusColor();
 
+
+    String placeholderColor();
+
+    /********************************************************************************************
+     *
+     * Command wizard
```
ashumilova: Category

ashumilova:
other OK

PR #511, opened 2015-12-11
CLDIDE-2644 added name to factory

mshaposhnik:


skabashnyuk:
ok. @evoevodin  @garagatyi @sleshchenko ?

platform-api/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryStore.java
```diff
@@ -54,13 +54,17 @@
...
+     * @param skipCount
+     *         skip items. Must be equals or greater then {@code 0}
      * @param  attributes
      *              - attribute pairs to search for
      *
      * @return - List {@code AdvancedFactoryUrl} if factory(s) exist and found, empty list otherwise
      * @throws org.eclipse.che.api.core.ApiException
      */
-    public List<Factory> findByAttribute(Pair<String, String>... attributes) throws ApiException;
+    public List<Factory> findByAttribute(int maxItems, int skipCount, Pair<String, String>... attributes) throws ApiException;
```
evoevodin: This method doesn't look good to me:
* Its main problem is in `Pair<String, String>...` see my [comment](https://github.com/codenvy/factory/pull/47#discussion_r47421195). 
It is not convenient to [use it](https://github.com/codenvy/che-core/blob/CLDIDE-2644/platform-api/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java#L384) + you have to [suppress](https://github.com/codenvy/che-core/blob/CLDIDE-2644/platform-api/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java#L370) warnings.
* Its javadoc worth nothing.
  * `Retrieve factory by given attribute name and value.` but method consumes array of attributes
  * `@throws org.eclipse.che.api.core.ApiException` general exception for all cases is bad practice + violates the [codestyle](https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#generally)
  * ` Must be equals or greater then {@code 0}` What if `skipCount` is negative ? How should implementation react on it? You defined just a half of the contract for `skipCount` parameter.

skabashnyuk:
ok

PR #512, opened 2015-12-11
Remove obsolete code related to disabling/enabling extensions loading

azatsarynnyy:


vparfonov:
ok

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/extension/ExtensionDescription.java
```diff
@@ -24,12 +24,11 @@
  * @author <a href="mailto:nzamosenchuk@exoplatform.com">Nikolay Zamosenchuk</a>
```
vzhukovskii: It would be good to remove author email. We agreed to remove such definition of author and leave only First and Last names

azatsarynnyy: It would be good to replace it in all our code base

vzhukovskii: Agree

PR #517, opened 2015-12-15
BUG FIX: result limit propagation in search operations

tareqhs:
1. Pass original request's maxItems through QueryExpression
2. Remove check of totalHits - the result is limited regardless

Currently, there is an incorrect check for TopDocs.totalHits that throws an exception if the total is greater than a hard-coded limit (1000). The limit that is given int the request is totally ignored when the actual search is ignored. Also, The maxItems should be propagated through the QueryExpression, and the check for totalHits should be totally removed since Lucene limits the number of result according to the parameters regardless of the total number of hits in the main index.


tareqhs:
@rimasirich
@Ori-Libhaber

PR #526, opened 2015-12-16
CLDIDE-2656 : Support both http and https in git server read-only URL

rainf0x:
This pull request is backport  of a commit  from master https://github.com/codenvy/che-core/commit/cc010b2901f1fa30192e7528d46d2735e27d505c


skabashnyuk:
ok

PR #527, opened 2015-12-16
IDEX-3632. Refine loader widget

RomanNikitenko:
Add the icons for the statuses of loader.
Change the icon for error state.
Fix bug: initialize workspace operation is never in Success state


vparfonov:
ok

PR #528, opened 2015-12-16
Move 'core' preference pages (Appearance, Plugins) to the same package where 'PreferencesPresenter' is

azatsarynnyy:
@vparfonov WDYT?


vparfonov:
ok

PR #529, opened 2015-12-16
Idex 3619

vparfonov:
Provide URL that will be used in the dashboard to load a given workspace into the IDE


skabashnyuk:
What this pr is doing. What is IDEX-3619? Can you update description?

skabashnyuk:
ok

PR #530, opened 2015-12-16
CLDIDE-2645 accept named factories

mshaposhnik:
- Removed unused legacy param;
- Added "validate" option to factory client;
- Fix find by attribute service


skabashnyuk:
ok

PR #532, opened 2015-12-17
IDEX-2552: add full-text search feature

svor:


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/CoreLocalizationConstant.java
```diff
@@ -565,4 +565,27 @@
...
     @Key("hot.keys.table.item.title")
     String hotKeysTableItemTitle();
+
+    @Key("action.full.text.search")
+    String actionFullTextSearch();
+
+    @Key("action.full.text.search.description")
+    String actionFullTextSearchDescription();
+
+    String search();
```
azatsarynnyy: I guess this method is unused
So it should be removed

vparfonov:
ok

PR #535, opened 2015-12-17
Mark readonly opened files with another color

vzhukovskii:
Feature request.
If user navigates in source code by pressing F4, he can open decompiled class from jar library, so we may mark editor tab with special color to allow user understand where exactly he is.

Screencast of this functionality https://slack-files.com/T02G3VAG4-F0GTN22GZ-354488aedd

@evidolob @vparfonov @skabashnyuk

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/widgets/editortab/EditorTabWidget.ui.xml
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/widgets/editortab/EditorTabWidget.ui.xml
@@ -27,6 +27,8 @@
         @eval activeEditorTabBackgroundColor org.eclipse.che.ide.api.theme.Style.theme.activeEditorTabBackgroundColor();
         @eval focusedEditorTabBackgroundColor org.eclipse.che.ide.api.theme.Style.theme.focusedEditorTabBackgroundColor();
         @eval focusedEditorTabBorderBottomColor org.eclipse.che.ide.api.theme.Style.theme.focusedEditorTabBorderBottomColor();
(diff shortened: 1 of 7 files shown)


azatsarynnyy:
really useful :+1:

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/widgets/editortab/EditorTabWidget.ui.xml
```diff
@@ -39,6 +41,13 @@
             color: tabTextColor;
             cursor: default;
             margin-bottom: 1px;
+
+            -webkit-transition: all 0.3s ease-in-out,
+                                fill 0.3s ease-in-out,
```
skabashnyuk: How it will look like on Spartan?

vzhukovskii: Need to check. Another question, do we officially support spartan?

benoitf: You may use transition property without prefix
http://caniuse.com/#feat=css-transitions

vzhukovskii: Below here is simple transition declaration:
```
transition: all 0.3s ease-in-out
```

And according to this reference http://www.w3schools.com/css/css3_transitions.asp it should

vzhukovskii: I've added this css effect only for smooth switching between editor tabs.

vparfonov:
ok

PR #539, opened 2015-12-18
Mark each Test folder child and Jar child with different background color

vzhukovskii:
Like in IDEA, see screenshots attached below:

Background
Background

@vparfonov @skabashnyuk @ashumilova


vparfonov:
ok

PR #542, opened 2015-12-19
IDEX-3612: Remove workspace from registry after workspace start error

evoevodin:
Implement [WorkspaceStatus](https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/workspace/WorkspaceStatus.java) spec by `RuntimeWorkspaceRegistry` which basically means the following: 
* Runtime workspaces won't stack in `STARTING` state if dev-machine failed to start
* Runtime workspace won't stack in `STOPPING` state if dev-machine failed to stop
* Runtime workspace status won't be affected if any machine different from dev failed to start
* Registry returns _up to date `RuntimeWorkspace` instance_ (with already started machines)

Cover `RuntimeWorkspaceRegistry` with tests.
Fix tomcat stop behaviour, now machines are stopped in `MachineManager#cleanup()` not in `RuntimeWorkspaceRegistry`

@skabashnyuk, @garagatyi, @akorneta could you please review?

### DIFF
--- a/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/RuntimeWorkspaceRegistry.java
+++ b/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/RuntimeWorkspaceRegistry.java
@@ -10,17 +10,17 @@
  *******************************************************************************/
 package org.eclipse.che.api.workspace.server;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ListMultimap;
 
-import org.eclipse.che.api.core.ApiException;
 import org.eclipse.che.api.core.BadRequestException;
 import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.ServerException;
-import org.eclipse.che.api.core.model.machine.Machine;
 import org.eclipse.che.api.core.model.machine.MachineConfig;
 import org.eclipse.che.api.core.model.workspace.Environment;
+import org.eclipse.che.api.core.model.workspace.EnvironmentState;
 import org.eclipse.che.api.core.model.workspace.RuntimeWorkspace;
 import org.eclipse.che.api.core.model.workspace.UsersWorkspace;
 import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
@@ -51,13 +51,21 @@ import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.STOPPING;
 /**
  * Defines {@link RuntimeWorkspace} internal API.
  *
- * <p>Workspaces are stored in memory - in 2 Maps.
- * First for <i>identifier -> workspace</i> mapping, second for <i>owner -> list of workspaces</i> mapping.
+ * <p>Registry implements {@link WorkspaceStatus} contract for runtime workspaces,
+ * the only implementation remark here is that in the case of {@link WorkspaceStatus#STOPPED}
+ * status registry removes workspace from in-memory storage instead of updating its status to STOPPED.
+ *
+ * <p>All methods which return instance of {@link RuntimeWorkspace} return the
+ * copy of real object, which means that instance modification won't affect real object.
+ * This is done for keeping data in actual state and for avoiding race-conditions.
+ * All operations performed by registry are synchronous.
+ *
+ * <p>The implementation is thread-safe.
+ * Workspaces are stored in memory - in 2 Maps. First for <i>identifier -> workspace</i> mapping,
+ * second for <i>owner -> list of workspaces</i> mapping(which speeds up fetching workspaces by owner).
  * Maps are guarded by {@link ReentrantReadWriteLock}.
  *
- * <p>It is thread-safe!
- *
- * @author Eugene Voevodin
+ * @author Yevhenii Voevodin
  * @author Alexander Garagatyi
  */
 @Singleton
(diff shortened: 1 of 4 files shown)


platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/RuntimeWorkspaceRegistry.java
```diff
@@ -334,54 +306,123 @@ public RuntimeWorkspaceImpl get(String workspaceId) throws NotFoundException {
...
+            throw ex;
         }
 
-        final List<MachineImpl> machines = new ArrayList<>();
+        // Updating workspace status and machines list
+        lock.writeLock().lock();
+        try {
+            addRunningMachine(devMachine);
+            final RuntimeWorkspaceImpl workspace = idToWorkspaces.get(workspaceId);
+            if (workspace != null) {
```
skabashnyuk: is this normal if workspace can be null?

evoevodin: Nice remark.

Workspace may be `null` if and only if `@PreDestroy` was performed and workspace was removed from in-memory storage, so it is the same to `if (isStopping)`, i missed that machine won't be stopped by `@PreDestroy` method as it stops all machines that were added to `RuntimeWorkspace` instance, but in this case `devMachine` hasn't been added yet, so it won't be destroyed. I am going to fix it, thank you.

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/RuntimeWorkspaceRegistry.java
```diff
@@ -272,7 +246,7 @@ public void update(RuntimeWorkspaceImpl update) throws ServerException, NotFound
      *         workspace identifier to perform check
      * @return true if workspace is running, otherwise false
      */
-    public boolean isRunning(String workspaceId) {
+    public boolean hasRuntime(String workspaceId) {
```
vzhukovskii: What "runtime" is in your vision? Why do you rename this method?

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/RuntimeWorkspaceRegistry.java
```diff
@@ -334,54 +306,123 @@ public RuntimeWorkspaceImpl get(String workspaceId) throws NotFoundException {
...
+        if (envName == null) {
+            throw new BadRequestException(format("Couldn't start workspace '%s', environment name is null", workspace.getName()));
+        }
+        final EnvironmentState environment = workspace.getEnvironments().get(envName);
+        if (environment == null) {
+            throw new BadRequestException(format("Couldn't start workspace '%s', workspace doesn't have environment '%s'",
+                                                 workspace.getName(),
+                                                 envName));
+        }
+        if (environment.getRecipe() != null && !"docker".equals(environment.getRecipe().getType())) {
```
vzhukovskii: Move "docker" hardcoded string to some shared constant, I'm sure that other components may have checking like this. It will help us with refactoring in future.

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -533,7 +531,7 @@ void stopWorkspaceAsync(String workspaceId) {
         executor.execute(() -> {
             try {
                 workspaceRegistry.stop(workspaceId);
-            } catch (ConflictException | NotFoundException | ServerException e) {
+            } catch (ConflictException | NotFoundException | ServerException  e) {
```
vzhukovskii: Do we need to format this line?

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/RuntimeWorkspaceRegistry.java
evoevodin: > What "runtime" is in your vision?

 its rather API vision than mine, if you are interested review [related models](https://github.com/codenvy/che-core/tree/master/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model).

> Why do you rename this method?

I did it because `isRunning` is confusing name and may be used wrong.
As you can see the name of the class is `RuntimeWorkspaceRegistry` which is responsible for managing `Runtime` representations of the `UsersWorkpace`.
In this case consider the following example:

``` 
if (workspaceRegistry.hasRuntime(workspaceId)) {
    // ...
}
```
Which actually means _If  workspace has runtime representation_. 
When: 

``` 
if (workspaceRegistry.isRunning(workspaceId)) {
    // ...
}

```
is confusing and may be considered as if _If workspace has runtime representation with  [RUNNING](https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/workspace/WorkspaceStatus.java#L59)  status_

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/RuntimeWorkspaceRegistry.java
evoevodin: Thank you for this remark, but moving to the constant is not __that__ solution.
In this case it is rather temporary check than permanent one, later this code will be replaced with environment management.

platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/RuntimeWorkspaceRegistry.java
vzhukovskii: I see. Sounds clear for me now, thx.

skabashnyuk:
ok

garagatyi:
щл

PR #548, opened 2015-12-21
CLDIDE-2654 : Add Authorization header between communication with API and remote runner. Add @RolesAllowed annotaions in SlaveBuilderService and SlaveRunnerService

rainf0x:
This is pull request for issue https://jira.codenvycorp.com/browse/CLDIDE-2654


platform-api/che-core-api-runner/src/main/java/org/eclipse/che/api/runner/RemoteRunnerProcess.java
```diff
@@ -161,4 +169,12 @@ private void doRequest(String url, String method, final OutputProvider output) t
             conn.disconnect();
         }
     }
+
+    private String getAuthenticationToken() {
```
skabashnyuk: you can inline this method

skabashnyuk:
Other ok

PR #552, opened 2015-12-21
IDEX-3876: Change the titles and messages of all Che notifications

vzhukovskii:
@vparfonov @ashumilova


vparfonov:
maybe better solution add message on client side, it allow  make localization

vzhukovskii:
Its a good remark, but it may estimation for this task is undefined. I think we should do this pluing by plugin in future.

skabashnyuk:
-1 to changing message in vfs. This work will need to be done again after we update our vfs implementation.

ashumilova:
+1

vparfonov:
ok

PR #553, opened 2015-12-21
IDEX-3910: Add ability to include in DTO's equals mechanism specific fields

vzhukovskii:
Description of problem provided in the issue: https://jira.codenvycorp.com/browse/IDEX-3910

@vparfonov @skabashnyuk WDYT?

### DIFF
--- a/platform-api/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImpl.java
+++ b/platform-api/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImpl.java
@@ -13,13 +13,13 @@ package org.eclipse.che.dto.generator;
 import org.eclipse.che.dto.shared.CompactJsonDto;
 import org.eclipse.che.dto.shared.DTO;
 import org.eclipse.che.dto.shared.DelegateTo;
+import org.eclipse.che.dto.shared.Compared;
 import org.eclipse.che.dto.shared.JsonFieldName;
 import org.eclipse.che.dto.shared.SerializationIndex;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonParser;
 import com.google.gson.stream.JsonWriter;
 
 import java.io.IOException;
@@ -32,6 +32,9 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+
+import static com.google.common.collect.Iterables.filter;
+import static com.google.common.collect.Lists.newArrayList;
 
 /** Abstract base class for the source generating template for a single DTO. */
 abstract class DtoImpl {
@@ -412,4 +415,17 @@ abstract class DtoImpl {
      * @return String representing the source definition for the DTO impl as an inner class.
      */
     abstract String serialize();
+
+    /**
+     * Filter input list of methods by finding {@link Compared} annotation.
+     *
+     * @param methods
+     *         method list
+     * @return filtered list
+     */
+    protected List<Method> getComparableMethods(List<Method> methods) {
+        return newArrayList(filter(methods, input -> {
+            return input.isAnnotationPresent(Compared.class);
+        }));
+    }
 }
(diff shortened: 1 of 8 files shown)


garagatyi:
So if at least one of the getters has annotation Comparable other getters won't be used in equals and hashcode generation.
Am I understand right?

skabashnyuk:
-1  so far. See comments in jira

vzhukovskii:
Yes. All methods, that are annotated with this annotation will participate in hashcode and equals generation. Otherwise if no one of the getters is annotated with this annotation, then behavior will be default: all fields will be included in hashcode and equal method.

skabashnyuk:
What real life problem you are trying to solve?

vzhukovskii:
For example in the project explorer I have Project Node that represent our Project Config Dto.
In some case user wants to reconfigure the projet. Okay, I have my Project Config Dto and I give this Dto to our configure dialog, then this dialog perform given work:

```java
    /** Open the project wizard for updating the given {@code project}. */
    public void show(@NotNull ProjectConfigDto project) {
        resetState();
        wizardMode = UPDATE;
        projectPath = project.getPath();
//creating new Dto! another question - why?
        final ProjectConfigDto dataObject = dtoFactory.createDto(ProjectConfigDto.class)
                                                      .withType(project.getType())
                                                      .withName(project.getName())
                                                      .withDescription(project.getDescription())
                                                      .withAttributes(new HashMap<>(project.getAttributes()));
        dataObject.setMixins(project.getMixins());
        showDialog(dataObject);
    }
```

Here we create new Dto for project configuration, but on this step I still can use equals to find my object. Its okay, but, when you click on "Config Project" button, configuration dialog will send copy of Dto to server to update the project. On this step magic begins, server side may provide into attributes various additional attributes (git mixins, artifact ids, etc...), it seems that all okay, but the project type may be still the same, the project path and name may be still also the same, so logically I got the same project in response. And then I need to update my node in Project Explorer with the same Dto.

So I have the new Dto after configuring. Logically new Dto is the same as old Dto, before configure (type, path, name) but it has different attributes, so in this case I couldn't use equal method. Thats the point.

Thats why I need to do smth like that:
```java
    @Override
    public void onProjectUpdated(ProjectUpdatedEvent event) {
        ProjectNode node = null;

//getting all nodes from project explorer
        for (Node seekNode : view.getAllNodes()) {
//and compare each node with path!
            if (seekNode instanceof ProjectNode &&
                ((ProjectNode)seekNode).getData().getPath().equals(event.getPath())) {
                node = (ProjectNode)seekNode;
                break;
            }
        }

//fetch new project config dto
        final ProjectConfigDto updatedDescriptor = event.getUpdatedProjectDescriptor();

//and update it in the node
        node.setData(updatedDescriptor);

//then refresh the view
        view.refresh(node);
//...
    }
```

but with this solution (PR) I can mark getPath() to be sure that only path will be include in equals method, so my code will reduces to this:

```java
        for (AbstractProjectBasedNode seekNode : view.getAllNodes()) {
//in this comparison only getPath() will be checked
//and I'm not related to it during this step
            if (seekNode.getData().equals(event.getUpdatedProjectDescriptor()) {
//in case if I'm sure that project path still the same after configuring
                   seekNode.setData(event.getUpdatedProjectDescriptor());
            }
        }
```
But on this moment I can't do this, because old Project Config Dto and new one has different in one or two additional attributes...

evoevodin:
-1 for this solution.

You are trying to solve the certain problem by changing DTO framework, which is obviously not good.
I am pretty sure that we have code that depends on the ProjectConfig state equality. 
And what you did, u just changed the way of hashCode and equals computation, don't you think that it will broke something?
In your example you need the only projects `path` but what if i need something else, what if i rely on project attributes as well as on the path?, should i mark `getAttributes` as `@Compared` ? If i mark it i will break the logic you did. 
So it is not the solution for the problem you are trying to solve. 

2 well known general points why not to do equals like you proposed:
* Equals is rather state equals than identity equals
* Pure `hashCode` may slow down hash-based collections performance

azatsarynnyy:
Few words about this comment: ```//creating new Dto! another question - why?```.
You can find answer to this question by reviewing history of this code. See how this block looks in 3.x branch. May be someone did refactoring and didn't understand purpose of this code block.

vparfonov:
+0 for me because this additional feature if you don't like it just don't use

vzhukovskii:
equals in java.io.File works in the common way:

```java
java.io.UnixFileSystem#compare

    public int compare(File f1, File f2) {
        return f1.getPath().compareTo(f2.getPath());
    }
```

Our ItemReference and ProjectConfig is representation of VirtualFile on server which is represent the real file on file system.

garagatyi:
-1 I don't think that this solution is good for whole DTO framework.
And I agree with @evoevodin that it can not solve situation when you need comparison by one field and I need by 2 extra fields.

vzhukovskii:
As for me its useless to compare object by all fields, for files and project configurations path will be anough

skabashnyuk:
You are mixing your business login with ideology of DTO as transport object. This change is potential risk  of very complex and unexpected issues.

PR #555, opened 2015-12-22
IDEX-3759: Fix parsing multiline command

azatsarynnyy:


commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/CommandLine.java
```diff
@@ -38,7 +38,7 @@ public CommandLine(String... args) {
 
     /** Create container for command line based on the specified {@code commandLine}. */
     public CommandLine(String commandLine) {
-        final String[] args = commandLine.split("\\s+");
```
vparfonov: What about multiple space "   "

azatsarynnyy: Yeah, need to improve. For multiple spaces as arguments separator it will return spaces as command's arguments.

PR #559, opened 2015-12-23
IDEX-3854 Update error code returned if there is an error in ServletProxy

benoitf:
error 404 changed into 503 if machine not here (service unavailable)
error 500 changed into 502 if there is error when communicating (bad gateway)

Change-Id: I0853af71062d7c102a8015c16c8f61097a9f8d2e


garagatyi:
ok for me

skabashnyuk:
ok

PR #560, opened 2015-12-23
IDEX-3841: Focus element handles key events only for the first create…

vzhukovskii:
…d tree widget in DOM

General stabilization of tree widget. And focus fixes. For now when user opens window with focus element it becomes active.

@vparfonov @svor


vparfonov:
ok

PR #561, opened 2015-12-23
IDEX-3691: manage deferred breakpoints

tolusha:
When JVM activate breakpoint then respective mark should be set by BreakpointManager


skabashnyuk:
ok. @vparfonov ?

vparfonov:
ok

PR #564, opened 2015-12-24
IDEX-3852: command selector improvements

olexii4:
@ashumilova WDYT?

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/action/Separator.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/action/Separator.java
@@ -13,8 +13,7 @@ package org.eclipse.che.ide.api.action;
 /**
  * Represents a separator.
  *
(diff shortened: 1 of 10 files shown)


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/action/Separator.java
```diff
@@ -22,6 +21,7 @@
     private String myText;
 
     public Separator() {
+        this(null);
```
sleshchenko: Can you explain why do we need this line?

vzhukovskii: I gues to initialize separator with null text

sleshchenko: I thought It is null by default :) But Oleksii told me that there is some problem with undefined and null value. Thanks for explanation.

vzhukovskii:
+1

PR #566, opened 2015-12-24
IDEX-3842: Fix GitHub Authorize application

vinokurig:
$wnd.IDE.config is a deprecated variable and it returns empty workspace name, so I check that loaded page url is equels to given url.
@vparfonov @skabashnyuk


skabashnyuk:
What value can be in path variable?

vinokurig:
Path variable receives path from  loaded URL in Authorization window.
When Authorization window start loading the path variable receives "blank", but when Authorization window loaded, user clicks "Authorize application" and it is successfully authorized, Authorize window loads given redirect URL. After that Path variable receives path from loaded redirect URL ("/ws" +  workspaceName). Checking of path == path of given redirect URL returns true and window is closing with successfully authorize response.

vparfonov:
do you test in on SDK?

vparfonov:
ok

skabashnyuk:
ok

PR #571, opened 2015-12-25
IDEX-3970: exclude recent files from action list

vzhukovskii:
https://jira.codenvycorp.com/browse/IDEX-3970

@vparfonov


vparfonov:
ok

PR #573, opened 2015-12-25
IDEX-3760: Fix import project after page refresh

dimasnurenko:
@vparfonov


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/FindReplaceAction.java
```diff
@@ -72,7 +72,8 @@ public void actionPerformed(ActionEvent event) {
                                                                                              .withReplace(replace)
                                                                                              .withReplacemode(mode)));
 
-        vfsServiceClient.replaceInCurrentWorkspace(appContext.getCurrentProject().getRootProject().getPath(),
+        vfsServiceClient.replaceInCurrentWorkspace(appContext.getWorkspaceId(),
```
sleshchenko: replace in current workspace but you should specify workspace id. Looks strange

vparfonov:
ok

PR #575, opened 2015-12-27
CLDIDE-2638 Save snapshot in onprem version

skabashnyuk:


platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/impl/InstanceKeyImpl.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Map based implementation of {@link org.eclipse.che.api.machine.server.spi.InstanceKey}.
+ *
+ * @author Sergii Kabashniuk
+ */
+
```
garagatyi: extra line

garagatyi:
ok

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/impl/InstanceKeyImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    public InstanceKeyImpl(InstanceKey instanceKey) {
+        this.fields = new HashMap<>(instanceKey.getFields());
+    }
+
+    /**
+     * @return unmodifiable copy of fields.
+     */
+    @Override
+    public Map<String, String> getFields() {
+        return Collections.unmodifiableMap(fields);
```
garagatyi: It is not possible to change field after instance creation. Why you create unmodifiable map each time this method is called? I would prefer use it once.

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -814,6 +814,13 @@ private SnapshotImpl doSaveMachine(SnapshotImpl snapshot, Instance machine) thro
 
             snapshotWithKey = new SnapshotImpl(snapshot);
             snapshotWithKey.setInstanceKey(machine.saveToSnapshot(machine.getOwner()));
+            //TODO remove old snapshot
```
garagatyi: Do we need that comment?

garagatyi:
ok

PR #578, opened 2015-12-28
IDEX-3911: Reduce the size of notification popup when there is no mes…

vzhukovskii:
…sage

@slemeur @vparfonov


ashumilova:
+1

vparfonov:
ok

PR #581, opened 2015-12-29
IDEX-3912: Move project template service to workspace master

dimasnurenko:
@vparfonov


platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/shared/dto/ProjectTemplateDescriptor.java
```diff
@@ -62,12 +57,15 @@
 
     ProjectTemplateDescriptor withDescription(String description);
 
-    //
-
     String getRecipe();
```
vparfonov: remove *recipe

vparfonov:
ok

PR #583, opened 2015-12-29
IDEX-2726: consider dev machine as running only after successful star…

garagatyi:
…t of ws agent
@skabashnyuk @evoevodin

### DIFF
--- a/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncher.java
+++ b/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncher.java
@@ -1,0 +1,23 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2015 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.machine.server;
+
+import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.machine.server.exception.MachineException;
+
+/**
+ * Starts ws agent in the machine and wait until ws agent sends notification about its start
+ *
+ * @author Alexander Garagatyi
+ */
+public interface WsAgentLauncher {
+    void startWsAgent(String workspaceId) throws NotFoundException, MachineException, InterruptedException;
+}
--- a/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncherImpl.java
+++ b/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncherImpl.java
@@ -1,0 +1,118 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2015 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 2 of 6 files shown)


skabashnyuk:
ок

akorneta:
ok

skabashnyuk:
ок

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncher.java
```diff
@@ -0,0 +1,23 @@
...
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.machine.server;
+
+import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.machine.server.exception.MachineException;
+
+/**
+ * Starts ws agent in the machine and wait until ws agent sends notification about its start
```
evoevodin: I think the sentence is not correct one, you should use eihter `Starts ws agent ... and waits ...` or `Start ws agent ... and  wait ...`

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncher.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @author Alexander Garagatyi
+ */
+public interface WsAgentLauncher {
+    void startWsAgent(String workspaceId) throws NotFoundException, MachineException, InterruptedException;
```
evoevodin: Is it necessary to have javadoc for this method?

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncherImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        this.wsAgentPingPath = apiEndpoint.getPath();
+    }
+
+    @Override
+    public void startWsAgent(String workspaceId) throws NotFoundException, MachineException, InterruptedException {
+        final Instance devMachine = machineManager.getDevMachine(workspaceId);
+        try {
+            machineManager.exec(devMachine.getId(),
+                                new CommandImpl(WS_AGENT_PROCESS_NAME, wsAgentStartCommandLine, "Arbitrary"),
+                                String.format(WS_AGENT_PROCESS_OUTPUT_CHANNEL, workspaceId));
```
evoevodin: I think in this case it is better to have `public static` method for getting channel.
Why:
The modifier of the `WS_AGENT_PROCESS_OUTPUT_CHANNEL` is public, which means that you want to reuse it in the other place, so you will do the same operation there `String.format(WS_AGENT_PROCESS_OUTPUT_CHANNEL, workspaceId)`.  If you create the method for it the usage will be obvious and won't let the developer to miss something

evoevodin:
Ok

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncher.java
garagatyi: yes

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncher.java
garagatyi: It is interface with one method, so javadoc for interface explains everything. Is it not enough in your opinion?

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncherImpl.java
garagatyi: +1

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/WsAgentLauncher.java
evoevodin: it is okay =)

PR #586, opened 2015-12-29
CLDIDE-2654 : Remove @RolesAllowed annotation on method /internal/runner/status/{runner:.*}/{id} in SlaveRunnerService

rainf0x:


skabashnyuk:
ок

PR #590, opened 2016-01-03
Authentication token fixes

tareksha:
Some fixes to make pass authentication token correctly across multi-threading in Che.
- Do not not add an explicit token query parameter is the download-sources URL that is passed from the build request from BuildQueue to the slave builder, leave it to the authentication header.
- Do not not add an explicit token query parameter is the URL of the deployment sources that is passed from the run request from RunQueue to the slave runner, leave it to the authentication header.
- Add authorization header to the HTTP requests issued from HTTP download plugin so that core components do not fail due to authorization issues when using it.
- Propagate thread-local context to default builder task in Che, so that the sources manager issues the download-sources HTTP call with the required authentication information available.


skabashnyuk:
ok. @vparfonov ? 
PR should be squashed before merge.

vparfonov:
ok

skabashnyuk:

PR #594, opened 2016-01-04
CLDIDE-2670: decouple workspace location finder from VFS

garagatyi:


skabashnyuk:
I would say we need to create interface with one method. Lets say String parent, String child.

commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/WorkspaceIdHashLocationFinder.java
```diff
@@ -0,0 +1,52 @@
...
+package org.eclipse.che.commons.lang;
+
+import java.io.File;
+
+/**
+ * Tool that calculates workspace files location on filesystem by the hash of workspace id.
+ *
+ * <p>We can have a lot of workspace and create root folder for all of them at the same level of filesystem
+ * may be inefficient. Keep workspace root folder in some hierarchy tree, use hash code for it.
+ *
```
skabashnyuk: Can you describe in javadoc the tree structure

garagatyi: done

skabashnyuk:
ок

PR #595, opened 2016-01-04
IDEX-3924: Display preview URL in console panel

vzhukovskii:
@slemeur @vparfonov @ashumilova @skabashnyuk

### DIFF
--- a/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/CommandImpl.java
+++ b/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/CommandImpl.java
@@ -10,9 +10,9 @@
  *******************************************************************************/
 package org.eclipse.che.api.machine.server.model.impl;
 
-import org.eclipse.che.api.core.model.machine.Command;
-
 import java.util.Objects;
+
+import org.eclipse.che.api.core.model.machine.Command;
 
 /**
  * Data object for {@link Command}.
@@ -24,6 +24,7 @@ public class CommandImpl implements Command {
     private String name;
     private String commandLine;
     private String type;
+    private String previewUrl;
 
     public CommandImpl(String name, String commandLine, String type) {
         this.name = name;
@@ -35,6 +36,7 @@ public class CommandImpl implements Command {
         this.name = command.getName();
         this.commandLine = command.getCommandLine();
         this.type = command.getType();
+        this.previewUrl = command.getPreviewUrl();
     }
 
     @Override
@@ -65,6 +67,15 @@ public class CommandImpl implements Command {
     }
 
     @Override
+    public String getPreviewUrl() {
+        return previewUrl;
+    }
+
+    public void setPreviewUrl(String previewUrl) {
+        this.previewUrl = previewUrl;
+    }
+
+    @Override
     public boolean equals(Object obj) {
         if (this == obj) {
             return true;
@@ -75,7 +86,8 @@ public class CommandImpl implements Command {
         final CommandImpl command = (CommandImpl)obj;
         return Objects.equals(name, command.name) &&
                Objects.equals(commandLine, command.commandLine) &&
-               Objects.equals(type, command.type);
+               Objects.equals(type, command.type) &&
+               Objects.equals(previewUrl, command.previewUrl);
     }
 
     @Override
@@ -84,6 +96,7 @@ public class CommandImpl implements Command {
         hash = 31 * hash + Objects.hashCode(name);
         hash = 31 * hash + Objects.hashCode(commandLine);
         hash = 31 * hash + Objects.hashCode(type);
+        hash = 31 * hash + Objects.hashCode(previewUrl);
         return hash;
     }
 
@@ -93,6 +106,7 @@ public class CommandImpl implements Command {
                "name='" + name + '\'' +
                ", commandLine='" + commandLine + '\'' +
                ", type='" + type + '\'' +
+               ", previewUrl='" + previewUrl + '\'' +
                '}';
     }
 }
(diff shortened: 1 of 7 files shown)


platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/CommandImpl.java
```diff
@@ -65,26 +65,28 @@ public void setType(String type) {
...
+    public void setPreviewUrl(String previewUrl) {
+        this.previewUrl = previewUrl;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        CommandImpl command = (CommandImpl)o;
+        return com.google.common.base.Objects.equal(getName(), command.getName()) &&
```
azatsarynnyy: Why you use ```com.google.common.base.Objects``` instead of ```java.util.Objects```?

vzhukovskii: It's just autoreplacement during generating equals through the Intellij IDEA. Will fix that.

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/CommandImpl.java
```diff
@@ -65,26 +65,28 @@ public void setType(String type) {
...
+    public String getPreviewUrl() {
+        return previewUrl;
+    }
+
+    public void setPreviewUrl(String previewUrl) {
+        this.previewUrl = previewUrl;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
azatsarynnyy: This method's body isn't formatted

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/machine/Command.java
```diff
@@ -37,4 +37,9 @@
      * Returns command type (i.e. 'maven')
      */
     String getType();
+
+    /**
+     * Returns command preview URL. (i.e. 'http://awesome.com:8080/context')
+     */
+    String getPreviewUrl();
```
vparfonov: pls, double check with this changes in API model personally i don't like it. 
But we can accept it if necessary for 4.0-beta 
@skabashnyuk @garagatyi @gazarenkov  your opinion need here

TylerJewell:
1. Can we move the preview URL portion of the command popup to be the last entry, after "Command line"?  The command line is first in people's minds, then the URL is second.

2. In the display of the previewURL, can it be browser blue instead of green?  Otherwise ppl may not understand it is clickable.

3. Is the URL clickable?

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/model/machine/Command.java
vzhukovskii: I don't like either. But according to issue we need to be able to store preview url for each command.

garagatyi: Preview url as a string is useless. [Here](https://codenvy.slack.com/archives/general/p1451318006000395) Tyler proposed another scheme. If we are OK to implement this, we need also realize how to set file: preview object.
Preview object as  string won't allow us to implement proposed strategies.

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/CommandImpl.java
```diff
@@ -65,17 +67,23 @@ public void setType(String type) {
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
evoevodin: Violates [Lisokv substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) from _Effective java 2nd ed item 8_

vzhukovskii: It's an artifact after generation equals method via intellij idea, reverted to original comparison algorithm.

platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/CommandImpl.java
```diff
@@ -73,9 +84,10 @@ public boolean equals(Object obj) {
             return false;
         }
         final CommandImpl command = (CommandImpl)obj;
-        return Objects.equals(name, command.name) &&
-               Objects.equals(commandLine, command.commandLine) &&
-               Objects.equals(type, command.type);
+        return Objects.equals(getName(), command.getName()) &&
```
evoevodin: Previous version of the equals was pretty ok, you could just add a new line of `Objects.equals` and that's it, why did you add getters here?

vzhukovskii: Haven't looked on it. Fixed, thanks.

vzhukovskii:
@TylerJewell here you go


>Can we move the preview URL portion of the command popup to be the last entry, after "Command line"? The command line is first in people's minds, then the URL is second.
Done

>In the display of the previewURL, can it be browser blue instead of green? Otherwise ppl may not understand it is clickable.
Color has been changed

>Is the URL clickable?
Yes, it's simple anchor element

vzhukovskii:
Updated state:

PR #596, opened 2016-01-05
IDEX-4000 Change email to alias in UserService

skabashnyuk:
In method find I've replaced query parameter email with query parameter alias.
It will call the same UserDao method findByAlias.

### DIFF
--- a/platform-api-client-gwt/che-core-client-gwt-user/src/main/java/org/eclipse/che/api/user/gwt/client/UserServiceClientImpl.java
+++ b/platform-api-client-gwt/che-core-client-gwt-user/src/main/java/org/eclipse/che/api/user/gwt/client/UserServiceClientImpl.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package org.eclipse.che.api.user.gwt.client;
 
+import com.google.inject.Inject;
+
 import org.eclipse.che.api.user.shared.dto.UserDescriptor;
 import org.eclipse.che.ide.MimeType;
 import org.eclipse.che.ide.rest.AsyncRequestCallback;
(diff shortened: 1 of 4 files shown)


platform-api-client-gwt/che-core-client-gwt-user/src/main/java/org/eclipse/che/api/user/gwt/client/UserServiceClientImpl.java
```diff
@@ -97,8 +97,8 @@ public void getUserById(@NotNull String id, AsyncRequestCallback<UserDescriptor>
 
     /** {@inheritDoc} */
     @Override
-    public void getUserByEmail(@NotNull String email, AsyncRequestCallback<UserDescriptor> callback) {
-        String requestUrl = FIND + "?email=" + email;
+    public void getUserByAlias(@NotNull String alias, AsyncRequestCallback<UserDescriptor> callback) {
+        String requestUrl = FIND + "?alias=" + alias;
```
evoevodin: if alias is `null` it will be added to the params, which means that user will get a message _"User with alias `null` doesn't exist"_ when he should get kind of  _"'alias' query parameter is required"_, you should add alias only if it is not null.

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -280,12 +280,12 @@ public UserDescriptor getById(@ApiParam(value = "User ID") @PathParam("id") Stri
     @GenerateLink(rel = LINK_REL_GET_USER_BY_EMAIL)
     @RolesAllowed({"user", "system/admin", "system/manager"})
     @Produces(APPLICATION_JSON)
-    public UserDescriptor getByEmail(@ApiParam(value = "User email", required = true) @QueryParam("email") @Required String email,
+    public UserDescriptor getByAlias(@ApiParam(value = "User alias", required = true) @QueryParam("alias") @Required String alias,
                                      @Context SecurityContext context) throws NotFoundException, ServerException, ConflictException {
-        if (email == null) {
-            throw new ConflictException("Missed parameter email");
+        if (alias == null) {
+            throw new ConflictException("Missed parameter alias");
```
evoevodin: I understand that all you did is variable rename, but anyway don't you think that changing of this code to throw `BadRequestException`  would improve this service, or maybe you can change all the methods of the service to throw `BadRequestExcetpion` in the case of not valid data.

skabashnyuk: I`m agree with you. But this is out of scope of this task

platform-api-client-gwt/che-core-client-gwt-user/src/main/java/org/eclipse/che/api/user/gwt/client/UserServiceClientImpl.java
skabashnyuk: I`m agree with you. But this is out of scope of this task

PR #597, opened 2016-01-05
Extended support for module paths

tareqhs:
Current code perform a simple text comparison to check if the module being conflicts with an existing module. Fixed this by comparing actual Path objects.

Current code does not update the modules list if a non-project parent folder of the module is renamed - it explicit requires the renamed folder to be a project folder in order to perform any module updates. Fixed.

Current code does not update projects other than the project that contains the module folder - it only cuts the renamed folder's path and checks that the enclosing project has the renamed folder listed as a module. Fixed.


skabashnyuk:
@vparfonov ?

vparfonov:
ok

skabashnyuk:
ок

skabashnyuk:

PR #599, opened 2016-01-05
IDEX-3510: Project Explorer refresh button does not work and is not s…

vzhukovskii:
…hown if Projects tree is empty


vparfonov:
ok

PR #600, opened 2016-01-05
Add special property for IDE url context

vparfonov:
we need it in property becouse context can be fifferent depend on packaging for example in Che it will be 'ide' but in hosted - 'ws'


vparfonov:
@skabashnyuk @evoevodin

skabashnyuk:
ok

PR #602, opened 2016-01-05
IDEX-3576: Add git remote and branch to projects attributes

vinokurig:
@vparfonov @skabashnyuk @sleshchenko please review

### DIFF
--- a/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectType.java
+++ b/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectType.java
@@ -21,12 +21,18 @@ import javax.inject.Singleton;
 @Singleton
 public class GitProjectType extends TransientMixin {
 
-    public static final String VCS_PROVIDER_NAME = "vcs.provider.name";
+    public static final String VCS_PROVIDER_NAME       = "vcs.provider.name";
+    public static final String GIT_CURRENT_BRANCH_NAME = "git.current.branch.name";
+    public static final String GIT_REPOSITORY_REMOTES  = "git.repository.remotes";
 
     @Inject
     public GitProjectType(GitValueProviderFactory gitRepositoryValueProviderFactory) {
         super("git", "git");
         addVariableDefinition(VCS_PROVIDER_NAME, "Is this git repo or not?", false,
+                              gitRepositoryValueProviderFactory);
+        addVariableDefinition(GIT_CURRENT_BRANCH_NAME, "Name of current git branch", false,
+                              gitRepositoryValueProviderFactory);
+        addVariableDefinition(GIT_REPOSITORY_REMOTES, "List of git repository remote addresses", false,
                               gitRepositoryValueProviderFactory);
     }
 }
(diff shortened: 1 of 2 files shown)


vparfonov:
ok

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectType.java
```diff
@@ -22,11 +22,17 @@
 public class GitProjectType extends TransientMixin {
 
     public static final String VCS_PROVIDER_NAME = "vcs.provider.name";
+    public static final String CURRENT_BRANCH_NAME = "current.branch.name";
+    public static final String REPOSITORY_REMOTES = "repository.remotes";
```
vzhukovskii: Formatter

evoevodin: Actually this variant is also [possible](https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#fields-alignment), but agree alligment is preferable.

vinokurig: done

skabashnyuk:
ok

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectType.java
```diff
@@ -21,12 +21,18 @@
...
+    public static final String VCS_PROVIDER_NAME   = "vcs.provider.name";
+    public static final String CURRENT_BRANCH_NAME = "current.branch.name";
+    public static final String REPOSITORY_REMOTES  = "repository.remotes";
 
     @Inject
     public GitProjectType(GitValueProviderFactory gitRepositoryValueProviderFactory) {
         super("git", "git");
         addVariableDefinition(VCS_PROVIDER_NAME, "Is this git repo or not?", false,
                               gitRepositoryValueProviderFactory);
+        addVariableDefinition(CURRENT_BRANCH_NAME, "Name of current brunch", false,
```
sleshchenko: Fix "brunch"

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectType.java
```diff
@@ -21,12 +21,18 @@
 @Singleton
 public class GitProjectType extends TransientMixin {
 
-    public static final String VCS_PROVIDER_NAME = "vcs.provider.name";
+    public static final String VCS_PROVIDER_NAME   = "vcs.provider.name";
+    public static final String CURRENT_BRANCH_NAME = "current.branch.name";
```
sleshchenko: Should we use commons name of these variables for all vcs? If yes maybe it would be better to add vcs. as prefix of attributes name. If no please think about adding of git. prefix to names of these (CURRENT_BRANCH_NAME and REPOSITORY_REMOTES) attributes

sleshchenko:
other OK

PR #606, opened 2016-01-06
IDEX-4001 add export workspace menu

mshaposhnik:


skabashnyuk:
ok

skabashnyuk:
@vparfonov ?
@evoevodin ?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/ExportConfigAction.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public void updateInPerspective(@NotNull ActionEvent event) {
+        CurrentProject activeProject = appContext.getCurrentProject();
+        event.getPresentation().setEnabled(activeProject != null);
+    }
+
+    /*
+    * Trick to force browser download the file instead of opening json on new tab
+    */
+    private native void downloadFile(String url, String wsName)/*-{
```
benoitf: Why we have to hack on client side ?
Why server is not providing the file content as a download ?

skabashnyuk: no it's not

benoitf: on 3.x factory export
we had header for that
https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java#L573-L575

$ curl -i https://codenvy.com/api/factory/workspacewdlne95weroleom2/MovieFun-TomEE
HTTP/1.1 200 OK
Server: Apache-Coyote/1.1
Set-Cookie: JSESSIONID=6092FDFBC6539F5D5618A53FF686D1FC; Path=/api/; Secure; HttpOnly
ETag: "d78afc0ba5324533b75d40c325928bf4"
Content-Disposition: attachment; filename=MovieFun-TomEE.json
Cache-Control: public, no-cache, no-store, no-transform
Content-Type: application/json
Content-Length: 929
Date: Wed, 06 Jan 2016 10:22:19 GMT

skabashnyuk: I know. But workspace api is not a factory API and we think it's not logical to add that header to workspace API method.

benoitf: why not add then an export method providing json as a download ?
it would be natural for every clients to view it as a json download

skabashnyuk: Because it's a client problem not api. We are not willing to add content-disposition to any GET method.

benoitf: "export of a workspace" is a server side operation that provides a file.
so this is not a client problem

slemeur: Please provide an API for that. All CHE clients be must be able use it. We defined requirements for that.

benoitf: note: maybe we could add a query parameter or header parameter in the request so we receive it as a download ?
current method is kept but we can change the return behavior by sending metadata to the server

skabashnyuk: Workspace API provide information to do that. We think it's consistent enough  to do his job.

benoitf: consistent : It means tons of hack on the IDE, UD and CLI clients to be able to download the file in browser client ?

skabashnyuk: https://jira.codenvycorp.com/browse/IDEX-4019
Is that ok for you?

benoitf: yes I think it adresses all issues

benoitf: thanks

PR #615, opened 2016-01-12
Idex 4003 -  Rename some parameters of workspace Environment model

mshaposhnik:
- defaultEnvName -> defaultEnv
- memorySize -> ram
- environment -> From Map to Array


benoitf:
this PR needs to be synchronized with UD as well

skabashnyuk:
ok. @vparfonov ?

vparfonov:
ok

PR #616, opened 2016-01-12
IDEX-4019 Introduce a new filter to provide download of the result

benoitf:
Change-Id: Ied519730b6cbab055305f13285874d8bf28dc703

### DIFF
--- a/platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/EverrestDownloadFileResponseFilter.java
+++ b/platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/EverrestDownloadFileResponseFilter.java
@@ -1,0 +1,57 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.everrest;
+
+import org.eclipse.che.api.core.rest.DownloadFileResponseFilter;
+import org.everrest.core.ApplicationContext;
+import org.everrest.core.Filter;
+import org.everrest.core.GenericContainerResponse;
+import org.everrest.core.ResponseFilter;
+import org.everrest.core.impl.ApplicationContextImpl;
+
+import javax.ws.rs.core.Request;
+import javax.ws.rs.core.Response;
+
+import static javax.ws.rs.core.HttpHeaders.CONTENT_DISPOSITION;
+
+/**
+ * JAX-RS implementation of download filter.
+ * @author Florent Benoit
+ */
+@Filter
+public class EverrestDownloadFileResponseFilter extends DownloadFileResponseFilter implements ResponseFilter {
+
+    /**
+     * Filter the given container response.
+     *
+     * @param containerResponse
+     *         the reponse to use
+     */
+    public void doFilter(GenericContainerResponse containerResponse) {
+        containerResponse.getResponse();
+
+        // Get the request
+        ApplicationContext applicationContext = ApplicationContextImpl.getCurrent();
+        Request request = applicationContext.getRequest();
+
+        // Apply header if all if correct
+        String filename = getFileName(request, containerResponse.getContentType(), applicationContext, containerResponse.getStatus());
+        if (filename != null) {
+            if (hasCompliantEntity(containerResponse.getEntity())) {
+                // it has been changed, so send response with updated header
+                Response.ResponseBuilder responseBuilder =
+                        Response.fromResponse(containerResponse.getResponse()).header(CONTENT_DISPOSITION, "attachment; filename=" + filename);
+                containerResponse.setResponse(responseBuilder.build());
+            }
+        }
+
+    }
+}
--- a/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DownloadFileResponseFilter.java
+++ b/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DownloadFileResponseFilter.java
@@ -1,0 +1,124 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.core.rest;
+
+import org.eclipse.che.dto.server.JsonSerializable;
+
+import javax.ws.rs.HttpMethod;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.Request;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriInfo;
+import java.util.List;
+
+import static org.eclipse.che.api.core.rest.DownloadFileResponseFilter.EntityType.JSON_SERIALIZABLE;
+import static org.eclipse.che.api.core.rest.DownloadFileResponseFilter.EntityType.STRING;
+import static org.eclipse.che.api.core.rest.DownloadFileResponseFilter.EntityType.UNKNOWN;
+
+/**
+ * Abstract Filter used to provide the json result as a file download operation.
+ * It is applying on GET method and JSON content type only.
+ *
+ * @author Florent Benoit
+ */
+public abstract class DownloadFileResponseFilter {
+
+    /**
+     * Entity type that we will be able to handle.
+     */
+    public enum EntityType {
+        JSON_SERIALIZABLE,
+        STRING,
+        UNKNOWN
+    }
+    /**
+     * Query parameter used to ask to specify headers that will propose JSON object to be downloaded.
+     */
+    public static final String QUERY_DOWNLOAD_PARAMETER = "downloadAsFile";
+
+
+    /**
+     * Check if we need to apply a filter
+     * @param request
+     * @return
+     */
+    protected String getFileName(Request request, MediaType mediaType, UriInfo uriInfo, int responseStatus) {
+
+        // manage only GET requests
+        if (!HttpMethod.GET.equals(request.getMethod())) {
+            return null;
+        }
+
+        // manage only OK code
+        if (Response.Status.OK.getStatusCode() != responseStatus) {
+            return null;
+        }
+
+        // Only handle JSON content
+        if (!MediaType.APPLICATION_JSON_TYPE.equals(mediaType)) {
+            return null;
+        }
+
+        // check if parameter filename is given
+        MultivaluedMap<String, String> queryParameters = uriInfo.getQueryParameters();
+        return queryParameters.getFirst(QUERY_DOWNLOAD_PARAMETER);
+    }
+
+    /**
+     * Check if entity is compliant with our filter
+     * @param entity the entity embedded in response.
+     * @return true if it's a type that we can handle
+     */
+    protected boolean hasCompliantEntity(Object entity) {
+        // no entity, skip
+        if (entity == null) {
+            return false;
+        }
+
+        // Check entity type
+        if (entity instanceof List) {
+            List<?> entities = (List)entity;
+            for (Object simpleEntity : entities) {
+                if (getElementType(simpleEntity) == UNKNOWN) {
+                    return false;
+                }
+            }
+        } else if (getElementType(entity) == UNKNOWN) {
+            // unknown entity type, will not configure it as a download
+            return false;
+        }
+
+        return true;
+    }
+
+
+    /**
+     * Helper method for getting the type of the JSON entity
+     *
+     * @param entity
+     *         the entity object
+     * @return the type of the element
+     */
+    protected EntityType getElementType(Object entity) {
+        if (JsonSerializable.class.isAssignableFrom(entity.getClass())) {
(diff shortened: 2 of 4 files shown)


platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @author Florent Benoit
+ */
+@Filter
+public class DownloadFileResponseFilter implements ResponseFilter {
+
+    /**
+     * Query parameter used to ask to specify headers that will propose JSON object to be downloaded.
+     */
+    private static final String QUERY_DOWNLOAD_PARAMETER = "downloadAsFilename";
```
sleshchenko: Name of this query parameter looks strange for me. Looks that I can download something as file name. Maybe just 'downloadAsFile' or something like 'targetFileName' would be better.

benoitf: ok tell me what is the final name of the parameter that you want

sleshchenko: :) I don't know. I just try find it.

benoitf: problem with downloadAsFile is that you expect parameter is a boolean not the name of the file (?downloadAsFile=foo.json)

targetFilename=dummy.json
downloadWithFilename=bar.json
responseDownloadFilename=foo.json

?

sleshchenko: I understand your point but downloadAsFile=[name of target file] Looks OK for me.

benoitf: PR updated to downloadAsFile

skabashnyuk: what about simple "?filename=blalba"?

vzhukovskii: or simple `?file=blabla`

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
```diff
@@ -0,0 +1,132 @@
...
+import javax.ws.rs.core.Request;
+import javax.ws.rs.core.Response;
+
+import java.util.List;
+
+import static javax.ws.rs.core.HttpHeaders.CONTENT_DISPOSITION;
+import static org.eclipse.che.everrest.DownloadFileResponseFilter.EntityType.JSON_SERIALIZABLE;
+import static org.eclipse.che.everrest.DownloadFileResponseFilter.EntityType.STRING;
+import static org.eclipse.che.everrest.DownloadFileResponseFilter.EntityType.UNKNOWN;
+/**
```
garagatyi: Pls  add new line here

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
benoitf: ok guys, please vote and then give me final answer

sleshchenko: file and filename are common names of parameters and they can override parameters from another api. I vote for downloadAsFile

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
```diff
@@ -0,0 +1,132 @@
...
+        // get entity of the response
+        Object entity = containerResponse.getEntity();
+
+        // no entity, skip
+        if (entity == null) {
+            return;
+        }
+
+        // Only handle JSON content
+        if (!MediaType.APPLICATION_JSON_TYPE.equals(containerResponse.getContentType())) {
```
garagatyi: I would like to make this check before retrieving entity, WDYT?

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
benoitf: I've applied formatter. ok

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
```diff
@@ -0,0 +1,132 @@
...
+            return;
+        }
+
+        // Get the request
+        ApplicationContext applicationContext = ApplicationContextImpl.getCurrent();
+        Request request = applicationContext.getRequest();
+
+        // manage only GET requests
+        if (!HttpMethod.GET.equals(request.getMethod())) {
+            return;
```
garagatyi: I would like to make this check before retrieving entity, WDYT?

skabashnyuk:
You would say this filter should not be connected to everrest.

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
benoitf: I'm +0 as we have a lot of JSON media type in JAX-RS BTW

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
benoitf: yep

benoitf:
I can use standard https://docs.oracle.com/javaee/7/api/javax/ws/rs/container/ContainerResponseFilter.html#filter-javax.ws.rs.container.ContainerRequestContext-javax.ws.rs.container.ContainerResponseContext- without problems but is it supported by everrest ?

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
benoitf: updated

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
benoitf: updated

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
benoitf: Also I didn't use something as simple ?file as it may collapse with classic parameters of some JAX-RS endpoints.

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        // Check entity type
+        if (entity instanceof List) {
+            List<?> entities = (List)entity;
+            for (Object simpleEntity : entities) {
+                if (getElementType(simpleEntity) == UNKNOWN) {
+                    return;
+                }
+                ;
```
garagatyi: Why you added empty operator here?

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+        } else if (getElementType(entity) == UNKNOWN) {
+            // unknown entity type, will not configure it as a download
+            return;
+        }
+
+        // update response if status is ok
+        if (Response.Status.OK.getStatusCode() == containerResponse.getResponse().getStatus()) {
```
garagatyi: I would like to make this check before retrieving entity, WDYT?

garagatyi:
+1 to use NOT everrest filter if it is possible.

skabashnyuk:
javax.ws.rs or javax.servlet Filter

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
benoitf: fixed

benoitf:
AFAIK everrest is not handling JAX-RS standard filters : https://github.com/codenvy/everrest/blob/9643bc4fee027e3664c6417679b9a29d239382ba/everrest-core/src/main/java/org/everrest/core/impl/RestComponentResolver.java

and this is not a Servlet feature which is too high in the stack (not want to operate on images, etc)

as long as you say to me that everrest is JAX-RS compliant with JAX-RS filters I will be happy to provide JAX-RS filter.
https://docs.oracle.com/javaee/7/api/javax/ws/rs/container/ContainerResponseFilter.html#filter-javax.ws.rs.container.ContainerRequestContext-javax.ws.rs.container.ContainerResponseContext-

skabashnyuk:
why do you think this is somehow related to everrest? May be in future we would like to change either rest framework either add this this feature on servlet. I would say javax.servlet.Filter is high enough.

benoitf:
ok so now I have provided a compliant JAX-RS filter and an everrest filter (because it is not working with standard filter)

benoitf:
it's ok then ?

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
```diff
@@ -0,0 +1,124 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.everrest;
+
```
skabashnyuk: may be we should change package to org.eclipse.che.rest?

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/DownloadFileResponseFilter.java
```diff
@@ -0,0 +1,124 @@
...
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.everrest;
```
skabashnyuk: this class should not be connected to everrest.

platform-api/che-core-api-core/src/main/java/org/eclipse/che/everrest/EverrestDownloadFileResponseFilter.java
```diff
@@ -0,0 +1,56 @@
...
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.everrest;
+
+import org.everrest.core.ApplicationContext;
+import org.everrest.core.Filter;
+import org.everrest.core.GenericContainerResponse;
+import org.everrest.core.ResponseFilter;
+import org.everrest.core.impl.ApplicationContextImpl;
```
skabashnyuk: I think we should avoid to use everrest specific classes

benoitf:
ok, so generic JAX-RS classes have been moved out of everrest package. Good cacth
I think now it's ok
Either everrest is used : Filter is available
Either another JAX-RS compliant rest framework is used : then the JAX-RS filter is available

benoitf:
Can I merge it so I can enable 'export workspace' feature in UD ?

skabashnyuk:
Why you are using this classes
+import org.everrest.core.ApplicationContext;
+import org.everrest.core.Filter;
+import org.everrest.core.GenericContainerResponse;
+import org.everrest.core.ResponseFilter;
+import org.everrest.core.impl.ApplicationContextImpl;

not javax.ws.rs?

skabashnyuk:
Can you remove EverrestDownloadFileResponseFilter?

benoitf:
in which class ? standard one or the one that is working in Everrest ?

benoitf:
If I remove this class, I can not enable it ! as everrest is not handling standard one

benoitf:
As I said, I have one pure JAX-RS class but everrest is not handling it.
Then I've a dedicated everrest class so I can have something working today as che is using everrest implementation.

When you decide to switch everrest to a JAX-RS compliant implementation or when everrest will be compliant, then we will be able to switch to the JAX-RS implementation

skabashnyuk:
lookslike ContainerResponseFilter is javax.ws.rs 2.0 and it's not implemented. going back to javax.servlet.Filter I would like to extrimly awoid everrest specific features if possible
Can we do  javax.servlet.Filter?

benoitf:
No, I don't want to use javax.servlet.filter and I already explained why.
Solution is fully expecting the associated JIRA issue
so I don't see why it can not be merged.

And the solution is working as I also provide two implementations so even if you drop everrest to another JAX-RS solution we can switch the implementation so there is no issue at all.

skabashnyuk:
ok

benoitf:
also maybe I could introduce JAX-RS 2.0 compliant filtering in everrest to remove dedicated filter

PR #619, opened 2016-01-13
IDEX-4053: Add attributes <String, String> field to CommandDTO and re…

vzhukovskii:
…move previewURL field from it

@vparfonov @gazarenkov @skabashnyuk @ashumilova pls, review


skabashnyuk:
ok @vparfonov ?

garagatyi:
equals, hashcode and toString are broken. Use getter for attributes. It will work in a different way before/after lazy initialization

vzhukovskii:
@garagatyi done

vparfonov:
ok

PR #622, opened 2016-01-14
IDEX-4093: Update tabs title

vzhukovskii:
@vparfonov @ashumilova 
this need to visual identify tabs if we use find text and usages in the same time


ashumilova:
+1

PR #623, opened 2016-01-14
IDEX-4028: Refine notification popups on light theme

vzhukovskii:
How it looks on Dark Theme

The same on the Light Theme

@ashumilova


ashumilova:
Ok

PR #625, opened 2016-01-14
Fix WS agent pinging

garagatyi:
Throw exception if pinging was unsuccessful and there is no time to ping more.
Refactor tests of ws agent pinging.
Add debug log.
@skabashnyuk


skabashnyuk:
ок

PR #635, opened 2016-01-15
Remove deprecated methods. …

vparfonov:
Add dummy impl for AnalyticsEventLogger in case we don't need analitics feature for Che int this moment


vparfonov:
@tolusha ?

tolusha:
ok

PR #642, opened 2016-01-16
IDEX-4101 Ignore not existing resources in class path during scanning classes

skabashnyuk:
Created reflections configuration builder that ignore resources that is not exists.
It happens when ide generators runs on generate-sources stage and ../target/classes folder is listed in class path but not exists.
Reflections print NPE for such folders. See more https://github.com/ronmamo/reflections/issues/111


vparfonov:
ok

PR #644, opened 2016-01-16
IDEX-4099: Fixed situation when IDE removes file's content on opening file in editor; IDEX-4076: Fix inability to type any content after creating a text file without extension

azatsarynnyy:
* Fixed situation when IDE removes file's content on opening file in editor.
* Fix inability to type any content after creating a text file without extension.

@vparfonov


vparfonov:
ok

PR #646, opened 2016-01-18
Remove unused property maven.model.version in pom.xml since we don't use org.apache.maven.maven-model anymore

azatsarynnyy:
Removed unused property **maven.model.version** in pom.xml since we don't use **org.apache.maven.maven-model** anymore
@vparfonov


skabashnyuk:
ok

PR #651, opened 2016-01-18
IDEX-3928: Add file filter to Git log command

vinokurig:
@vparfonov  @skabashnyuk please review

### DIFF
--- a/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/commands/LogCommand.java
+++ b/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/commands/LogCommand.java
@@ -26,8 +26,9 @@ import java.util.List;
  */
 public class LogCommand extends GitCommand<List<Revision>> {
 
-    private int    count;
-    private String branch;
+    private int          count;
+    private String       branch;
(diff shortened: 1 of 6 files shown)


che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/commands/LogCommand.java
```diff
@@ -46,6 +47,9 @@ public LogCommand(File place) {
         if (count > 0) {
             commandLine.add("-" + count);
         }
+        if (fileFilter != null && !fileFilter.isEmpty()) {
```
sleshchenko: Please take a look implementation of CommandLine https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/CommandLine.java#L62-L67
I think we can miss these checkings

vinokurig: done

skabashnyuk:
ok for me. @vparfonov ?

vparfonov:
ok

PR #658, opened 2016-01-20
Allow creating a DTO from JsonElement directly

tareqhs:
avoid an unnecessary double serialization if a JsonElement is availableand a DTO needs to be created from it.


skabashnyuk:
ok @vparfonov ?

vparfonov:
ok

skabashnyuk:

PR #659, opened 2016-01-21
CDEC-481: Provide REST API for getting list of all users from on-prem

tolusha:


skabashnyuk:
Not sure it should be part of dao. @gazarenkov  @vparfonov ?

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/dao/UserDao.java
```diff
@@ -117,4 +118,19 @@
...
+     *         the page number
+     * @param perPage
+     *         the the number entities per page
+     * @return list of users POJO or empty list if no users were found
+     * @throws BadRequestException
+     *         when page or perPage is negative
+     * @throws ServerException
+     *         when any other error occurs
+     */
+    List<User> getAll(int page, int perPage) throws ServerException, BadRequestException;
```
skabashnyuk: Our standard paging has format
 int maxItems, int skipCount

tolusha: List usually is used in such cases.

gazarenkov: For me OK to keep it the same DAO since it is a layer on the same database

skabashnyuk:
ok

tolusha:
What about /findall as REST method name? @gazarenkov

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -424,6 +427,40 @@ public UserDescriptor getByName(@ApiParam(value = "User email")
...
+    @Path("/findall")
+    @GenerateLink(rel = "get all users")
+    @RolesAllowed({"system/admin", "system/manager"})
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Get all users",
+                  notes = "Get all users in the system. Roles allowed: system/admin, system/manager.")
+    @ApiResponses({@ApiResponse(code = 200, message = "OK"),
+                   @ApiResponse(code = 400, message = "Bad Request"),
+                   @ApiResponse(code = 500, message = "Internal Server Error")})
+    public List<UserDescriptor> getAll(@ApiParam(value = "Max items") @QueryParam("maxItems") @DefaultValue("30") int maxItems,
```
skabashnyuk: @gazarenkov  wdyt about method signature ?

gazarenkov: It should be another Service with path like
/user/admin
also we should avoid using verbs in REST methods (resources) addressing 
so
@GET
 +    @Path("/findall") - not good

In this case I would say we do not need a Path at all.

PR #660, opened 2016-01-21
IDEX-2389: Add getProjectExplorerSelection method in SelectionAgent

vinokurig:
When performing git commit with "add selected files" check-box selected and console part is active, NPE appears because the getSelection method in Selection agent returned a selection of console part, but we need a selection of project explorer
@vparfonov @skabashnyuk  please review

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/parts/PartPresenter.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/parts/PartPresenter.java
@@ -32,7 +32,7 @@ import java.util.List;
 public interface PartPresenter extends Presenter {
     /** The property id for <code>getTitle</code>, <code>getTitleImage</code> and <code>getTitleToolTip</code>. */
     int TITLE_PROPERTY     = 0x001;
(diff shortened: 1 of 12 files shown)


skabashnyuk:
ok for me. @vparfonov ?

azatsarynnyy:
I'm disagree with this solution. It breaks concept of the **Selection agent**.
Each part has it's own object that **Selection agent** interacts with, some kind of part's context. Part can set any object into selection. **Selection agent** provides an access to the selection of the active part. 
This solution does not fit into this concept of the **Selection agent**.

Why don't you use ```projectExplorer.getSelection()``` if you need selection of the Project Explorer?

@vparfonov WDYT?

vinokurig:
I can, but I think old get selection method In Selection agent can be confusing as it can return anything. I added this method to split Selection agent, so no we can receive a selection from active part and a selection from project explorer instead of injecting ProjectExplorer just to get a selection from it.

azatsarynnyy:
This is the main idea of the **Selection Agent** - provide a way of getting current app-wide selection. It may be anything

vparfonov:
agree with @azatsarynnyy this not good decisions add getProjectExplorerSelection

PR #671, opened 2016-01-26
IDEX-3936: Move limit propagation in vfs from rework_vfs_4 branch

akorneta:


platform-api/che-core-api-vfs/src/main/java/org/eclipse/che/api/vfs/server/search/LuceneSearcher.java
```diff
@@ -126,6 +128,38 @@ public synchronized IndexWriter getIndexWriter() {
...
+            final String[] result = new String[topDocs.scoreDocs.length];
+            for (int i = 0, length = result.length; i < length; i++) {
+                result[i] = luceneSearcher.doc(topDocs.scoreDocs[i].doc).getField("path").stringValue();
+            }
+            return result;
+        } catch (IOException e) {
+            throw new ServerException(e.getMessage(), e);
+        } finally {
+            try {
+                searcherManager.release(luceneSearcher);
```
garagatyi: NPE can be thrown here

skabashnyuk:
ок. @vparfonov ?

vparfonov:
ok

PR #672, opened 2016-01-26
improve Machine API

garagatyi:
Add links
Return 400 if required parameter is missing instead of 403

@skabashnyuk @evoevodin


skabashnyuk:
ok

PR #675, opened 2016-01-27
CDEC-481: Provide REST API for getting list of all users from on-prem

tolusha:


platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/utils/UserTransformer.java
```diff
@@ -0,0 +1,129 @@
...
+import static org.eclipse.che.api.user.server.Constants.LINK_REL_GET_USER_BY_ID;
+import static org.eclipse.che.api.user.server.Constants.LINK_REL_GET_USER_PROFILE_BY_ID;
+import static org.eclipse.che.api.user.server.Constants.LINK_REL_REMOVE_USER_BY_ID;
+import static org.eclipse.che.api.user.server.Constants.LINK_REL_UPDATE_PASSWORD;
+
+/**
+ * @author Anatoliy Bazko
+ */
+public class UserTransformer {
+
```
skabashnyuk: Make this class similar to https://github.com/codenvy/che-core/blob/master/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/DtoConverter.java

tolusha:
@skabashnyuk @sleshchenko wdyt?

evoevodin:
:+1:

skabashnyuk:
ok

platform-api/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/LinksInjector.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Helps to inject {@link UserService} related links.
+ *
+ * @author Anatoliy Bazko
+ */
+public final class LinksInjector {
+
+    public static UserDescriptor injectLinks(UserDescriptor userDescriptor, ServiceContext serviceContext) {
+        final User currentUser = EnvironmentContext.getCurrent().getUser();
+        final UriBuilder uriBuilder = serviceContext.getBaseUriBuilder();
```
sleshchenko: WDYT about using getServiceUriBuilder() method from ServiceContext and remove all .path(UserProfileService.class)?

sleshchenko:
:+1:

PR #679, opened 2016-01-28
IDEX-4157. Use copy workspace object at operations with this one

RomanNikitenko:
This will allow don't use the same workspace object at the same time while different operations are executed.


vparfonov:
ok

PR #680, opened 2016-01-28
IDEX-4114: Ability to execute a command on context menu in project ex…

vzhukovskii:
…plorer

commands-context

@ashumilova @vparfonov @skabashnyuk 

Related PR: https://github.com/codenvy/che-plugins/pull/604


ashumilova:
+1

PR #691, opened 2016-02-03
Cldide 2659 - move  user resolving into separate class

mshaposhnik:


platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/UserResolver.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Resolves git user.
+ *
+ * @author Max Shaposhnik
+ *
+ */
+public interface UserResolver {
+
+    public GitUser getUser();
```
sleshchenko: You can remove 'public' for methods in interface

che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/LocalUserResolver.java
```diff
@@ -0,0 +1,69 @@
...
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static org.eclipse.che.dto.server.DtoFactory.newDto;
+
+/**
+ * Resolves user for git based on environment context.
+ *
+ * @author Max Shaposhnik
+ *
+ */
+public class LocalUserResolver implements UserResolver {
```
evoevodin: It seems that you missed `@Singleton` annotations as this class is stateless

che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/LocalUserResolver.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(LocalUserResolver.class);
+
+    private PreferenceDao preferenceDao;
```
evoevodin: make it `final`

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/UserResolver.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.eclipse.che.api.git.shared.GitUser;
+
+/**
+ * Resolves git user.
+ *
+ * @author Max Shaposhnik
+ *
+ */
+public interface UserResolver {
```
evoevodin: I think that this name should be scoped more, it may confuse people if they meet that interface.
I propose to rename this interface to `GitUserResolver` and add appropriate javadoc with reference to `GitUser`.

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/UserResolver.java
```diff
@@ -0,0 +1,25 @@
...
+ *******************************************************************************/
+package org.eclipse.che.api.git;
+
+import org.eclipse.che.api.git.shared.GitUser;
+
+/**
+ * Resolves git user.
+ *
+ * @author Max Shaposhnik
+ *
```
evoevodin: why do you add an empty line after the `@author` ?

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitUserResolver.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.eclipse.che.api.git.shared.GitUser;
+
+/**
+ * Resolves {@link GitUser} for any git related operations.
+ *
+ * @author Max Shaposhnik
+ */
+public interface GitUserResolver {
+
```
skabashnyuk: javadoc

skabashnyuk:
Other ok

PR #696, opened 2016-02-04
IDEX-3779: Fix reset cursor position in editor after switching perspective

dimasnurenko:
@vparfonov


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/parts/StorePartStateEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.ide.api.parts;
+
+import com.google.gwt.event.shared.EventHandler;
+import com.google.gwt.event.shared.GwtEvent;
+
+/**
+ * The class represents event wich is thrown when before we change perspective.
```
azatsarynnyy: > when before

_when_ or _before_? ))

dimasnurenko: )) i'll fix

vparfonov:
ok

PR #697, opened 2016-02-05
IDEX-4215 Updates icons so they are all compatible with Eclipse Licensing

vitaliy-guliy:


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/icon/FontAwesome.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.dom.client.Document;
+import com.google.gwt.dom.client.LinkElement;
+
+/**
+ * Font awesome style injector and set of predefined icons.
+ *
+ * @author Vitaliy Guliy
+ */
+public class FontAwesome {
```
azatsarynnyy: I think that this class is more related to the **che-core-ide-ui** module but not to the **che-core-ide-app**.
WDYT?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/client/BootstrapController.java
```diff
@@ -135,6 +136,8 @@ public void execute() {
 
     /** Displays the IDE */
     private void displayIDE() {
+        FontAwesome.injectCSS();
```
azatsarynnyy: It would be much more better to inject it through start-up [Component](https://github.com/codenvy/che-core/blob/master/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/core/Component.java) like [ZeroClipboardInjector](https://github.com/codenvy/che-core/blob/master/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/bootstrap/ZeroClipboardInjector.java). It will allow to avoid hardcode it there.

benoitf:
less and scss versions of FontAwesome should be removed (keep only css + fonts folder)
only one css should be kept (either minified or not minified version)

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/icon/FontAwesome.java
vparfonov: +1

vitaliy-guliy: Moved

vitaliy-guliy:
less, scss and css/font-awesome.css are removed

PR #707, opened 2016-04-24
Check permissions on symbolic links/relative paths

freedomp:
- No one has permissions to read/write files outside the workspace root
- Get children, delete and properties/acl related operations don't require
  permission on the target file in case of symbolic links.
  Other operations do.
- Added missing permission checks in case the file is a symbol link and che
  relied on the read permission being granted on the original file
- Add missing permission checks in get child and get children APIs
- Added support for copying/zipping symbolic links

Signed-off-by: Dror Cohen dror.cohen@sap.com

### DIFF
--- a/che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
+++ b/che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
@@ -75,6 +75,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -379,7 +380,7 @@ public class FSMountPoint implements MountPoint {
         if (!virtualFile.exists()) {
             throw new NotFoundException(String.format("Object '%s' does not exists. ", vfsPath));
         }
-        if (!hasPermission(virtualFile, BasicPermissions.READ.value(), true)) {
+        if (!hasPermission(virtualFile, BasicPermissions.READ.value(), true, false)) {
             throw new ForbiddenException(String.format("Unable get item '%s'. Operation not permitted. ", virtualFile.getPath()));
         }
         return virtualFile;
@@ -446,12 +447,19 @@ public class FSMountPoint implements MountPoint {
             return null;
         }
         final Path childPath = parent.getVirtualFilePath().newPath(name);
+        
+        // Find the last existing parent path (the actual parent of the new file) and check permissions on it
+        VirtualFileImpl actualParentFile = getLastExistingParent(parent, name);
+        if (!hasPermission(actualParentFile, BasicPermissions.READ.value(), true)) {
+            throw new ForbiddenException(String.format("Unable get item '%s'. Operation not permitted. ", childPath));
+        }
+        
         final VirtualFileImpl child = new VirtualFileImpl(new java.io.File(parent.getIoFile(), name), childPath, pathToId(childPath), this);
         if (child.exists()) {
             if (systemFilter.accept(workspaceId, child.getVirtualFilePath())) {
                 // Don't check permissions for file "misc.xml" in folder ".codenvy". Dirty huck :( but seems simplest solution for now.
                 // Need to work with 'misc.xml' independently to user.
-                if (!hasPermission(child, BasicPermissions.READ.value(), true)) {
+                if (!hasPermission(child, BasicPermissions.READ.value(), true, false)) {
                     throw new ForbiddenException(String.format("Unable get item '%s'. Operation not permitted. ", child.getPath()));
                 }
             }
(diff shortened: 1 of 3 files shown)


skabashnyuk:
Can you describe what for do you use symbolic links?

freedomp:
This is a security issue.
To prevent malicious use of symbolic links in a project, that point to files that the user is not allowed to read/write, such as outside his workspace, etc.

skabashnyuk:
How this link was made?

freedomp:
The link could have been made by a malicious user, inserted into his workspace.

skabashnyuk:
You mean over the rest API? I would like to understand the whole usecase that you are trying to solve.

seastranger:
There are several ways to create such a link - it can be created using the rest API or be imported through git or zip importers

skabashnyuk:
Interesting. Maybe this is a vulnerability and we should avoid to create it?

seastranger:
You mean not support symbolic links at all ? why not ?
it was vulnerability, but our pull request adds checks in all relevant places (hopefully) so that user can only access things inside his workspace he has permission to access. if someone attempts to import project containing such a link (access with no permission) the link file will just not be imported (created)

skabashnyuk:
In general I'm ok with this PR. Can you add some tests to demonstrate issue and fix.

tal-sapan:
Hi,

I would like to give an overview of this change:

**FSMountPoint**
- hasPermission method now receives another boolean parameter that determines, in case the file is a symbolic link, whether permissions on the target of the symbolic link should be followed (as well as permissions on the checked file). By default it's true.
- We don't allow to use regular methods to read/write files in the service dir (.vfs) because there are dedicated methods for this. If they are written via API we can't assure their content. This is now blocked in the hasPermission method.
- We don't allow to read/write files outside the current workspace. This is blocked in the hasPermission method.
- The methods that don't require the symbolic link's target to be checked are those that don't read/write the content of the target (the content of a folder is the list of files inside it): doGetVirtualFile, getChild, getChildren (the check on each child), delete, lock, updateACL, updateProperties.
- Since there are 2 different permission checks that can be performed now (follow/don't follow the symbolic link), more permission checks are required: methods that relied on the permission check to be done before the method was called now can't rely on it. These methods include getChildren, copy, getContent, zip.
- In some methods that do operations on a folder (read/write files), the permissions were not checked on the actual folder whose content is read/written, but only on the project level. Now we have to check the permission on the actual parent folder since it might be a symbolic link. The affected methods are getChild, createFile, createFolder.
- Deleting a symbolic link should not delete the target file, therefore we don't check permissions on children of symbolic links' children (and the target of symbolic link files) in doDelete.
- To prevent recursive copy of symbolic links, we now don't copy the content of symbolic link folders. The affected methods are doCopy, zip.
- Special handling is required for recursive symbolic links to prevent infinite loops/recursions. For this reason, we don't return the children of recursive symbolic links. The affected methods are getChild, doGetChildren.

**IoUtil**

When copying symbolic links we don't copy the content of the file, but instead create another symbolic link that points to the same target path (note that if the target path is relative this might not be the same file).

**FolderEntry**

The check if a folder is a project folder could now cause a ForbiddenException in case it's a symbolic link to a file the user doesn't have permission on. Since this use case doesn't necessarily mean there's a system error now, we don't throw an exception in this case but just return false (meaning the folder is not a project folder). This was done to support deleting a folder through the API.

skabashnyuk:
Ok for me. I think we good to go. @tareqhs  can you merge with changes?

che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
```diff
@@ -905,6 +968,20 @@ void updateContent(VirtualFileImpl virtualFile, InputStream content, String lock
         updateContent(virtualFile, null, content, lockToken, false);
     }
 
+    private boolean isPathUnderServiceDir(String pathStr)
+    {
```
benoitf, line 972: the { are not matching the current format of the class

che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
```diff
@@ -905,6 +968,20 @@ void updateContent(VirtualFileImpl virtualFile, InputStream content, String lock
...
+        java.nio.file.Path path = java.nio.file.Paths.get(pathStr);
+        Iterator<java.nio.file.Path> pathIterator = path.iterator();
+        while (pathIterator.hasNext()) {
+            java.nio.file.Path subPath = pathIterator.next();
+            if (subPath.getFileName().toString().equals(SERVICE_DIR))
```
benoitf, line 977: formatter is if () { on the same line

commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
```diff
@@ -390,6 +398,19 @@ private static void copy(File source, File target, FilenameFilter filter, boolea
...
+        if (target.exists() && !replaceIfExists) {
+            throw new IOException(String.format("File '%s' already exists. ", target.getAbsolutePath()));
+        }
+        Path newLink = target.toPath();
+        Path linkTarget = Files.readSymbolicLink(source.toPath());
+        if (replaceIfExists) {
+            Files.deleteIfExists(newLink);
+        }
+        Files.createSymbolicLink(newLink, linkTarget);
+        return;
```
benoitf, line 411: why return ?

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/FolderEntry.java
```diff
@@ -229,12 +229,13 @@ public FolderEntry createFolder(String name) throws ConflictException, ServerExc
      *         if an error occurs
      */
     public boolean isProjectFolder() throws ServerException {
-        final VirtualFileEntry projectFile;
+        VirtualFileEntry projectFile;
         try {
             projectFile = getChild(Constants.CODENVY_PROJECT_FILE_RELATIVE_PATH);
         } catch (ForbiddenException e) {
```
benoitf, line 235: we may move there the return projectFile != null && projectFile.isFile(); below

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/FolderEntry.java
```diff
@@ -229,12 +229,13 @@ public FolderEntry createFolder(String name) throws ConflictException, ServerExc
...
-            // If have access to the project then must have access to its meta-information. If don't have access then treat that as server error.
-            throw new ServerException(e.getServiceError());
+            // Assume that it's not a project if we don't have access to the project file.
+            // If it were a project we should have access to its meta-information because we have access to its folder.
+            projectFile = null;
```
benoitf, line 238: we may return false directly

skabashnyuk:
@freedomp @tareqhs  do you need this pr? Can you comment @benoitf questions?

PR #708, opened 2016-05-01
Fix missing message.properties file + remove .git folder if operation fails

blankg:


skabashnyuk:
ok. @vinokurig ?

che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -583,13 +595,32 @@ public void init(InitRequest request) throws GitException {
...
+            }
+
             throw new GitException(e.getMessage(), e);
         }
     }
 
+    private void deleteRepositoryFolder() {
+        try {
+            if (repository.getDirectory().exists()) {
+                FileUtils.delete(repository.getDirectory(), FileUtils.RECURSIVE | FileUtils.IGNORE_ERRORS | FileUtils.IGNORE_ERRORS);
```
vinokurig: why do you use `FileUtils.IGNORE_ERRORS` twice?

blankg:
Removed the duplicate IGNORE_ERRORS flag

vinokurig:
ok for me

che-core-git-impl-jgit/src/main/resources/org.eclipse.che.git.impl.jgit/messages.properties
```diff
@@ -0,0 +1,34 @@
+#
```
vinokurig: Properties file pthname is invalid, it should be: `che-core-git-impl-jgit/src/main/resources/org/eclipse/che/git/impl/jgit/messages.properties`

blankg:
properties path should work now

PR #710, opened 2016-05-04
Push to Gerrit add missing method for set and configData

offer8:
Add new set config API for Git implementation

Change-Id: Ie55b7d50bc211d6a23757ae7cc40e1c220a94247

Signed-off-by: Offer Shostak offershostak@gmail.com

### DIFF
--- a/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
+++ b/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
@@ -68,6 +68,7 @@ import javax.inject.Inject;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
@@ -379,20 +380,21 @@ public class GitService {
 
 
     @Path("config")
-    @POST
+    @GET
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
-    public Map<String, String> getConfig(ConfigRequest request) throws ApiException {
+    public Map<String, String> getConfig(@QueryParam("requestedConfig") List<String> requestedConfig)
+            throws ApiException {
         Map<String, String> result = new HashMap<>();
         try (GitConnection gitConnection = getGitConnection()) {
             Config config = gitConnection.getConfig();
-            if (request.isGetAll()) {
+            if (requestedConfig == null || requestedConfig.isEmpty()) {
                 for (String row : config.getList()) {
                     String[] keyValues = row.split("=", 2);
                     result.put(keyValues[0], keyValues[1]);
                 }
             } else {
-                for (String entry : request.getConfigEntry()) {
+                for (String entry : requestedConfig) {
                     try {
                         String value = config.get(entry);
                         result.put(entry, value);
@@ -403,6 +405,25 @@ public class GitService {
             }
         }
         return result;
+    }
+
+
+    @Path("config")
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    public void setConfig(ConfigRequest request) throws ApiException {
+        try (GitConnection gitConnection = getGitConnection()) {
+            Config config = gitConnection.getConfig();
+            for (Map.Entry<String, String> configData : request.getConfigEntries().entrySet()) {
+                try {
+                    config.set(configData.getKey(), configData.getValue());
+                } catch (GitException exception) {
+                    final String msg = "Cannot write to config file";
+                    LOG.error(msg, exception);
+                    throw new GitException(msg);
+                }
+            }
+        }
     }
 
     @Path("tag-list")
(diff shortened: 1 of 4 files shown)


offer8:
@tareqhs

skabashnyuk:
@vinokurig ?

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -79,10 +79,7 @@
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.net.URISyntaxException;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
vinokurig: Do not use `*` in imports, https://github.com/codenvy/codenvy-dev-resources/wiki/Source-file#imports

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -391,6 +388,17 @@ public void tagDelete(TagDeleteRequest request) throws ApiException {
                     String[] keyValues = row.split("=", 2);
                     result.put(keyValues[0], keyValues[1]);
                 }
+            } else if (request.isSet()) {
```
vinokurig: I think you should create a separate get api method with query parameter for this. It looks very strange doing `config.set` in method that called `getConfig`.

offer8:
Hi,

Thanks for your feedback,
Regarding import - I will provide a fix.
Regarding the "config.set" - what about adding new api called "setConfig"?

Regards,
Offer.

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
vinokurig: Please rework current `getConfig` method to a post api method with query parameter and add a new `getConfig` method as get api method with `ConfigRequest` DTO.

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -391,6 +388,17 @@ public void tagDelete(TagDeleteRequest request) throws ApiException {
...
+                for (Map.Entry<String, String> configData : request.getConfigData().entrySet()) {
```
vinokurig: Please rework current method to a GET api method with query parameter like this:
`@GET
Map<String, String> getConfigs(@QueryParam List<String> requestedConfigs) {
//returns all if requestedConfigs.isEmpty()
...
}`

and add a new `setConfig` PUT api method with `ConfigRequest` dto like this: 
`@PUT
void setConfig(ConfigRequest config) {
...
}`.

The ConfigRequest DTO should contain only information map of config entries to update, like this:
`@DTO
public interface ConfigRequest {
    Map<String, String> getConfigEntries();
    void setConfigEntries(Map<String, String> configEntries);
    ConfigRequest withConfigEntries(Map<String, String> configEntries);
}`

offer8:
Hi,

I fixed the code according to your instructions.

Regards,
Offer.

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -379,20 +380,21 @@ public void tagDelete(TagDeleteRequest request) throws ApiException {
 
 
     @Path("config")
-    @POST
+    @GET
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
-    public Map<String, String> getConfig(ConfigRequest request) throws ApiException {
+    public Map<String, String> getConfigs(@QueryParam("requestedConfigs") List<String> requestedConfigs)
```
vinokurig: Please rename `getConfigs` to `getConfig` and `requestedConfigs` to `requestedConfig`

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -405,6 +407,29 @@ public void tagDelete(TagDeleteRequest request) throws ApiException {
         return result;
     }
 
+
+    @Path("config")
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Map<String, String> setConfig(ConfigRequest request) throws ApiException {
```
vinokurig: It is not necessary to return a map because it returns the same that we requested. It would be better to make a `void` method

platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/shared/ConfigRequest.java
```diff
@@ -13,23 +13,18 @@
...
  * @author andrew00x
  */
 @DTO
 public interface ConfigRequest {
 
-    boolean isGetAll();
+    Map<String, String> getConfigEntries();
 
-    void setGetAll(boolean geAll);
+    void setConfigEntries(Map<String, String> ConfigEntries);
```
vinokurig: Please rename `ConfigEntries` to `configEntries`.

vinokurig:
Also please adopt `getConfig` method here: https://github.com/codenvy/che-core/blob/5c0c82c9defd2c08f52ee72126481f833c8718be/platform-api-client-gwt/che-core-client-gwt-git/src/main/java/org/eclipse/che/api/git/gwt/client/GitServiceClientImpl.java#L231-L238 to support our client side.

offer8:
Hi,

I pushed new code.
@vinokurig can you please take a look?

Regards,
Offer.

vinokurig:
OK

skabashnyuk:

PR #711, opened 2016-05-09
Improve error messages

blankg:


skabashnyuk:
@vinokurig ?

che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -418,8 +419,24 @@ public void clone(CloneRequest request) throws GitException, UnauthorizedExcepti
                 deleteRepositoryFolder();
             }
 
-            throw new GitException(exception.getMessage());
+            String message = generateSSLExceptionMessage(exception);
+            throw new GitException(message, exception);
+        }
+    }
+
+    private String generateSSLExceptionMessage(Throwable e) {
```
vinokurig: I think the name of the method is not clear enough. It says that method should return 'SSLException' message but given exception can be caused not only by an `SSLHandshakeException`

che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -418,8 +419,24 @@ public void clone(CloneRequest request) throws GitException, UnauthorizedExcepti
...
+        String message = e.getMessage();
+
+        //if e caused by an SSLHandshakeException - replace thrown message with a hardcoded message
+        while (e.getCause() != null) {
```
vinokurig: `e.getCause()` is used three times, may be it would be better to introduce a local variabe

blankg:
Regarding the method name generateSSLExceptionMessage, I can change it to a more generic name generateExceptionMessage

vinokurig:
in this case this method should be everywhere where `exception.getMessage()` is used. How about `getSpecialMessageIfSSLExceptionWasTrown`?

sleshchenko:
getSpecialMessageIfSSLExceptionWasTrown looks bad for me. Because this method return message even when SSLException was not thrown

blankg:
if we have a generic method it can later be extended to generate messages to other root causes and not just SSL error

sleshchenko:
@blankg +1 but in fact we don't generate this message. So just `getErrorMessage` or 'retrieveErrorMessage' look OK for me

PR #714, opened 2016-05-16
checkout API fix conflict exception - change GitException to APIexception

i053322:
checkout API change GitException to APIexception

Change-Id: I7bd58baa5d0e015107343f8ed1610ffea43b0a0b

Signed-off-by: i053322 yossi.balan@sap.com

### DIFF
--- a/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
+++ b/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/NativeGitConnection.java
@@ -10,6 +10,9 @@
  *******************************************************************************/
 package org.eclipse.che.git.impl.nativegit;
 
+
+import com.google.common.base.Strings;
+import org.eclipse.che.api.core.ApiException;
 import org.eclipse.che.api.core.UnauthorizedException;
 import org.eclipse.che.api.core.util.LineConsumerFactory;
 import org.eclipse.che.api.git.Config;
@@ -160,7 +163,7 @@ public class NativeGitConnection implements GitConnection {
     }
 
     @Override
-    public void cloneWithSparseCheckout(String directory, String remoteUrl, String branch) throws GitException, UnauthorizedException {
+    public void cloneWithSparseCheckout(String directory, String remoteUrl, String branch) throws ApiException, UnauthorizedException {
         /*
         Does following sequence of Git commands:
         $ git init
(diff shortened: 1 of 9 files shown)


tareksha:
why?

offer8:
The GitException type does not expose the right checkout conflict exception in the inheritance tree which we need to the full checkout scenario.

offer8:
Can someone please merge this one?

skabashnyuk:
Not yet. Can you explain more why you need that

i053322:
In checkout scenario when you move between branches you could get conflict error.
It could occurred:
1. if you did any change and then you to move to different branch that based on different remote.
   1. if you did any commit and you tried to move to different branch on the same remote 

In this case the error from the jcheckoutCommand.call() is CheckoutConflictException and not GitAPIException.

vinokurig:
Can you additionally catch CheckoutConflictException and handle it?

i053322:
This is what I did in my change I added another catch for the CheckoutConflictException

vinokurig:
Where? As I see you just changed throwing GitException to ApiException. As I understood, you want to handle CheckoutConflictException, but native git does not produce such exception. May be you should handle it in JgitConnection?

i053322:
I added the implementation for the catch  in the jgitConnection

vinokurig:
I can not see any changes in JgitConnection in this Pull request, I only see changes in NativeGitConnection

offer8:
Hi,

As first phase we change the che-api layer to support the new exception type.
Second phase, after the merge of JGit branch to 3.x is completed we will push the new logic which we inserted in the JGitConnection checkout method.

The reason for the change in the exception type is that we need to throw ConflictException which being extended from ApiException.
We do this in order to return the right HTTP status and message for the client.
This logic is placed in ApiExceptionMapper.java class that builds the relevant HTTP response according to the exception type.
GitException is not in the inheritance tree of APIException.

Regards,
Offer.

sleshchenko:
I agree that it would be better to throw exception with right response code. But I think that replacing GitException with ApiException is a wrong way to implement it. 
I think it would be better to remove GitException at all (or mark deprecated for now) and add to throws section specific child classes of ApiException to wit ServerException, ConflictException, NotFoundException etc. And add to java doc  of GitConnection information about required behaivour for implementation (which exception should be thrown in which cases). We do the same for our Daos interfaces for example [WorkspaceDao](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/WorkspaceDao.java#L58-66)

@skabashnyuk WDYT?

i053322:
i think you right and you proposal should be the way to raise error message in git.
For example in the checkout Jgit throw the following errors :
GitAPIException, RefAlreadyExistsException, RefNotFoundException, InvalidRefNameException, CheckoutConflictException
so in the JGitConnection in the checkout I need to create new Exception for each type meaning 
GitAPIException - throw ApiException 
RefAlreadyExistsException - need to create new message
RefNotFoundException,  - throw NotFoundException
InvalidRefNameException,  - need to create new message
CheckoutConflictException - throw ConflictException

This mean that I need also to add new check for the logic in the ApiExceptionMapper for the new message ( RefAlreadyExistsException , InvalidRefNameException )
Is it ok ?
Also what do you think I need to do in the native can I throw ApiExcption for all the messages ?
Can we start to do the change only for the checkout API and continue later for the other apis ?

Regards,
Yossi

gazarenkov:
I think the best way to apply everywhere is to:
- have fine grained logical exception on business logic
- add a mapping for these exceptions - to - REST friendly (ApiException) in a specific module 

so if I understand you well, I am +1 for this

i053322:
what do I need to do in the case of the native ? can I throw  ApiExcption ?

PR #715, opened 2016-05-16
Enhance BuilderSelectionStrategy with environment and request

sfroee:
This will extend the functionality of the Builder with a more generic way to select a builder out of available builders.

basically the change is to allow the builder selection strategy to consider the build request as well as the remote builder properties.

This is needed for running Che in Cloud Foundry environment, where there can be several available builders across different spaces.

Currently Che allows registering builders by the access criteria
"builderServerAccessCriteria": {
    "project": "",
    "workspace": ""
 }

and by matching this criteria together with default builder on project type the match is done. 
Than builder selection strategy implementation is called (if there are more then a single available builder) and the selection can filter by properties in the [RemoteBuilder](ttps://github.com/codenvy/che-core/blob/5c0c82c9defd2c08f52ee72126481f833c8718be/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteBuilder.java).

Also the getEnvironment() method on the Builder class is not accessible in RemoteBuilder , so it can't select using this data

This is not enough and hence the motivation for contribution.

The suggested  enhance do the following:
- class RemoteBuilder it saves the environment provided by the builder and expose it via get method.
- interface BuilderSelectionStrategy adds a default method that receives additional parameter of the build request, there the extra information can be used to better filter the builders with data coming from the Build request. the default implementation is used so not to affect already existing strategies 
- class BuildQueue - calls the builder strategy with the build request. 
  And also calls selector when there is a single builder available as it might not pass the selection criteria.

@tareqhs - please review

happy for any input 
Regards
Roy

### DIFF
--- a/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/BuilderSelectionStrategy.java
+++ b/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/BuilderSelectionStrategy.java
@@ -11,15 +11,34 @@
 package org.eclipse.che.api.builder;
 
 import java.util.List;
+import org.eclipse.che.api.builder.dto.BaseBuilderRequest;
 
 /**
- * Selects the 'best' SlaveBuilder from the List according to implementation. BuildQueue uses implementation of this interface fo
- * find the 'best' slave-builder for processing incoming build request. If more then one slave-builder available then BuildQueue collects
+ * Selects the 'best' SlaveBuilder from the List according to implementation. BuildQueue uses implementation of this interface to
+ * find the 'best' slave-builder for processing incoming build request. If slave-builder available then BuildQueue collects
  * them (their front-ends which are represented by SlaveBuilder) and passes to implementation of this interface. This implementation
  * should select the 'best' one.
  *
  * @author andrew00x
  */
 public interface BuilderSelectionStrategy {
+
+    /**
+     * Selects best matching {@code RemoteBuilder} that satisfies the implemented strategy
+     * @param slaveBuilders
+     * @return {@code RemoteBuilder} or {@code null} if there is no adequate selection
+     */
     RemoteBuilder select(List<RemoteBuilder> slaveBuilders);
+
+    /**
+     * Selects best matching {@code RemoteBuilder} that satisfies implemented strategy.
+     * Implemented strategy may consider both {@code BaseBuilderRequest} and {@code RemoteBuilder} properties
+     * This method is implementation optional, as selection by builder properties might be enough, and backward compatible.
+     * @param slaveBuilders
+     * @param request
+     * @return {@code RemoteBuilder} or {@code null} if there is no adequate selection
+     */
+    default RemoteBuilder select(List<RemoteBuilder> slaveBuilders, BaseBuilderRequest request){
+        return select(slaveBuilders);
+    }
 }
--- a/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteBuilder.java
+++ b/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteBuilder.java
@@ -27,10 +27,13 @@ import org.eclipse.che.api.core.rest.shared.Links;
 import org.eclipse.che.api.core.rest.shared.dto.Link;
 import org.eclipse.che.commons.lang.Pair;
 import org.eclipse.che.dto.server.DtoFactory;
+import org.eclipse.che.api.builder.dto.BuilderEnvironment;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Represents remote {@code Builder}.
@@ -57,6 +60,7 @@ public class RemoteBuilder {
     private final String     name;
     private final String     description;
     private final int        hashCode;
+    private final Map<String, BuilderEnvironment> environmentMap;
 
     private volatile long lastUsage = -1;
 
@@ -65,6 +69,7 @@ public class RemoteBuilder {
         this.baseUrl = baseUrl;
         name = builderDescriptor.getName();
         description = builderDescriptor.getDescription();
+        this.environmentMap = builderDescriptor.getEnvironments();
         this.links = new ArrayList<>(links);
         int hashCode = 7;
         hashCode = hashCode * 31 + baseUrl.hashCode();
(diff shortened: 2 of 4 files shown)


platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteBuilder.java
```diff
@@ -106,6 +110,14 @@ public long getLastUsageTime() {
     }
 
     /**
+     * Get Builder Environment map
+     * @return map of BuilderEnvironment
+     */
+    public Map<String, BuilderEnvironment> getBuilderEnvironment(){
+        return this.environmentMap;
```
tareksha: This should be wrapped with Collections.unmodifiableMap(map) to avoid unexpected modifications

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/BuilderSelectionStrategy.java
```diff
@@ -22,4 +23,8 @@
  */
 public interface BuilderSelectionStrategy {
     RemoteBuilder select(List<RemoteBuilder> slaveBuilders);
+
+    default RemoteBuilder select(List<RemoteBuilder> slaveBuilders, BaseBuilderRequest request){
+        return select(slaveBuilders);
+    }
```
tareksha, line 43: I suggest providing a default implementation for the shorter version, this way we require minimal implementation code for implementations that rely on either version of the method.
Currently there are only two builder selection strategies, changing them is one-liner ..
@skabashnyuk  agree ?

skabashnyuk:
Any chance to have a test here?

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteBuilder.java
sfroee: sure , will do

sfroee:
regarding unit tests , I'll see if I can add any

skabashnyuk:
Also @sfroee  can you please check if follow the rules of eclipse contribution. 
https://wiki.eclipse.org/Development_Resources/Contributing_via_Git

1. Sign eclipe CLA
2. Sign off git commit

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/BuilderSelectionStrategy.java
sfroee, line 43: The default implementation for the added method was for automatically support other strategy implementations, not from Che code. so other who took the che code and created their own strategy implementation will not break. if it's not a real case , then we can just update the single method of the interface and the current two implementations

sfroee, line 43: @skabashnyuk - what do you think about this point?

sfroee:
resolved comments from @skabashnyuk and @tareqhs 
please review

platform-api/che-core-api-builder/src/test/java/org/eclipse/che/api/builder/BuilderTest.java
```diff
@@ -95,6 +102,19 @@ public boolean removeListener(SourceManagerListener listener) {
                 }
             };
         }
+
+        @Override
+        public Map<String, BuilderEnvironment> getEnvironments() {
+            Map<String, BuilderEnvironment> env = new HashMap<>();
+
+            BuilderEnvironment builderEnvironment = DtoFactory.getInstance().createDto(BuilderEnvironment.class)
```
benoitf, line 110: you can use newDto(BuilderEnvironment.class)  (importing static newDto method from DtoFactory)

platform-api/che-core-api-builder/src/test/java/org/eclipse/che/api/builder/BuilderTest.java
```diff
@@ -95,6 +102,19 @@ public boolean removeListener(SourceManagerListener listener) {
...
+                    .withId("default")
+                    .withIsDefault(true)
+                    .withDisplayName(this.getName());
+
+            env.put(builderEnvironment.getId(), builderEnvironment);
+            return env;
```
benoitf: as we build a map with only one entry : 
return Collections.singletonMap(key, value)  (no need to build HashMap, put key, value and return it)

sfroee:
fixed remarks of @benoitf 
Thanks

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/BuilderSelectionStrategy.java
skabashnyuk, line 43: Changes should be backward compatible as much as possible. I think your variant is better. May be it make sense to add some javavdocs.

sfroee, line 43: Thanks @skabashnyuk 
I'll check how to add this and create another pull next week

skabashnyuk, line 43: Are  you going to close this pr?

sfroee, line 43: no, I'll override to this one , we want it in the code at the end

skabashnyuk, line 43: Javadoc in this PR is welcomed

sfroee:
Added java docs to BuilderSelectionStrategy per @skabashnyuk request
can it be merged ?

skabashnyuk:
ooooh come on. code is failing on license check ((

sfroee:
sorry for that , I'm checking how did this happen
should I create a new pull ?

skabashnyuk:
No. I've fixed it already https://github.com/codenvy/che-core/commit/81115b8fac9cb7487e5448942a56fef4f462f3f8

sfroee:
Thanks @skabashnyuk , found the root cause,
before working directly with github che repository, we develop it and test it in a local maintained che fork
there it is also validated with actual product tests.
unfortunately there the license was not correct (year not updated).
Later we cherry-pick to che core driven repository and there mistakenly the error propagated.

I'll check how to avoid this kind of errors for future contributions.
Thanks again
Roy

skabashnyuk:
ok. thanks

PR #719, opened 2016-06-21
Add "." to validity of project name .

i048695:
Dot (".") can be added to new project name.  

Change-Id: Ib79d741e70ac561b9bac4470661fe4f178195fb5
Signed-off-by: i048695 lev.zak@sap.com

### DIFF
--- a/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
+++ b/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
@@ -1579,7 +1579,7 @@ public class ProjectService extends Service {
         if (Strings.isNullOrEmpty(name)) {
             throw new BadRequestException("Project name required");
         }
-        if (!Pattern.compile("[a-zA-Z0-9]+[\\w-]*").matcher(name).matches()) {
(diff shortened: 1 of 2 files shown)


skabashnyuk:
PR is not clear. Can you elaborate a bit more about your changes? what is it and why do we need it?

i048695:
Hi
My name Lev Zak
I am from develop team of DEVX SAP.
I have task to enable create new project with dot(".") in name . I add some change to ProjectService.java   (if (!Pattern.compile("[a-zA-Z0-9]+[\\.\\w-]*").matcher(name).matches()) {  )

i048695:
Hi
We need to create project with dot in name . We can't replace dot "." to something else . This is the reason that i add this change .

skabashnyuk:
Examples?
dots can be at beginning or end of the name?

i048695:
For example 
aaa.bbb

tareksha:
Hi @skabashnyuk 
as Lev explained, the goal is to support project names that contain dots - much like java packages. 
do you approve the change?

skabashnyuk:
Yes.  In genera @tareqhs  we are good to go.
BTW are you able to add some tests?

skabashnyuk:
ok

PR #722, opened 2016-07-20
Allow modifying the environment variables of a Builder command process

tareksha:
An extension might require setting/modifying the environment variables of the forked process that execute the command. Currently, the Che builder does not modify the environment of the built process at all and all the environment variables of the parent java process are passed as-is. Not only does this cause some issue due to possibly missing environment variables, but it also might pose a security issue due to exposing "too much" from the environment of the parent java process.

This change adds a an optional method that can be overridden to manipulate the environment variables that are set for the command process that is executed by Che builders. It also upgrades the codes slightly to use Java 8's new native Process timing APIs.

### DIFF
--- a/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/Builder.java
+++ b/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/Builder.java
@@ -21,12 +21,8 @@ import org.eclipse.che.api.core.ApiException;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.core.notification.EventSubscriber;
-import org.eclipse.che.api.core.util.Cancellable;
-import org.eclipse.che.api.core.util.CancellableProcessWrapper;
 import org.eclipse.che.api.core.util.CommandLine;
-import org.eclipse.che.api.core.util.ProcessUtil;
 import org.eclipse.che.api.core.util.StreamPump;
-import org.eclipse.che.api.core.util.Watchdog;
 import org.eclipse.che.commons.lang.IoUtil;
 import org.eclipse.che.commons.lang.concurrent.ThreadLocalPropagateContext;
 import org.eclipse.che.dto.server.DtoFactory;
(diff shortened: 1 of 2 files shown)


tareksha:
@sfroee
@skabashnyuk

skabashnyuk:
are you sure this is backward compatible change?

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/Builder.java
```diff
@@ -486,6 +487,15 @@ public void cancelled(Cancellable cancellable) {
         };
     }
 
+	/**
+	 * Modify the environment variables that are set for a command being executed.
+	 * 
+	 * @param environmentVars The environment variables of the command process.
+	 * @param configuration Connfiguration of the build task.
+	 */
+	protected void modifyCommandEnvironment(Map<String, String> environmentVars, BuilderConfiguration configuration) {
```
sfroee: good change! 
can we have a logger passed to this method as well?  so we can comment any change in the build log

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/Builder.java
```diff
@@ -444,6 +444,7 @@ public Boolean call() throws Exception {
                 try {
                     ProcessBuilder processBuilder = new ProcessBuilder().command(commandLine.toShellCommand()).directory(
                             configuration.getWorkDir()).redirectErrorStream(true);
+                    modifyCommandEnvironment(processBuilder.environment(), configuration);
                     Process process = processBuilder.start();
```
sfroee, line 392: processBuilder.start() - need to be wrapped by some timer , so we can know the actual time spent in the command execution. - can be in another future change

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/Builder.java
tareksha: yup, i'll add it @sfroee

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/Builder.java
tareksha, line 392: I suggest using java 8's built-in waitFor(time, timeUnit) instead of using ad-hoc polling code. this areas is a candidate for several poll requests I sugess

tareksha:
@skabashnyuk yes it is, the added method has a default implementation and does not affect existing builder implementations.

tareksha:
@skabashnyuk  @sfroee 
the change has been updated to allow even more extensibility, plus using java 8's new standard Process API

skabashnyuk:
ok. @vparfonov @evoevodin ?

skabashnyuk:
@tareqhs  you can merge this pr

PR #723, opened 2016-07-29
Make SourcesManager an injected singleton

tareksha:
Currently each builder instantiates its own SourcesManagerImpl, taking the responsibility for its lifecycle as well. This takes the SourcesManager implementation out of the dependency injection, and causes creating clean-up task scheduler although a single task is enough.

The change makes SourcesManagerImpl an injected singleton that all builders share.


skabashnyuk:
Looks ok for me. @evoevodin @vparfonov ?

skabashnyuk:
@tareqhs  Can you make sure that Che is starting correctly?

tareksha:
hi @skabashnyuk , @evoevodin , @vparfonov
tested, fixed a typo and removed unnecessary comments.

vparfonov:
ok

PR #724, opened 2016-07-31
A central class for all builder clean-up operations

tareksha:
Currently each builder runs own thread for cleaning expired tasks, although a single universal thread that cleans all expired tasks in all builder is more efficient. This change moves the logic from individual builders to a central that uses a single thread to clean all expired tasks.

Change-Id: I80df89ff897e34783dd9300b6caec46264dc0c4a

### DIFF
--- a/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/BuilderCleaner.java
+++ b/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/BuilderCleaner.java
@@ -1,0 +1,73 @@
+package org.eclipse.che.api.builder.internal;
+
+import java.io.File;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.PreDestroy;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.eclipse.che.commons.lang.IoUtil;
+import org.eclipse.che.commons.schedule.ScheduleRate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Manages clean-up operations for build tasks of local slave builders.
+ * 
+ * @author Tareq Sharafy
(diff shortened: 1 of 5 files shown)


platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/BuilderCleaner.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Inject
+    public BuilderCleaner(BuilderRegistry builders, @Named(Constants.KEEP_RESULT_TIME) int keepResultTime) {
+        this.builders = builders;
+        this.keepResultTimeMillis = TimeUnit.SECONDS.toMillis(keepResultTime);
+        this.cleanupScheduler = Executors.newSingleThreadScheduledExecutor(
+                new ThreadFactoryBuilder().setNameFormat("BuilderManagerScheduler-%d").setDaemon(true).build());
+    }
+
+    @PostConstruct
```
skabashnyuk: you can use https://github.com/codenvy/che-core/blob/3.x/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/ScheduleRate.java
docs
https://github.com/eclipse/che/tree/master/core/commons/che-core-commons-schedule

skabashnyuk:
@vparfonov  @evoevodin  wdyt?

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/BuilderCleaner.java
tareksha: @skabashnyuk seems very useful, thanks! but it needs a patch because the scheduling Runnables do not propagate the thread-local context.

tareksha:
@skabashnyuk @evoevodin now using commons-schedule

vparfonov:
ok

PR #725, opened 2016-07-31
Propagate thread-local context in missing areas in che-core

tareksha:
These areas fail if some thread locals are missing, mostly the user token information in EnvironmentContext

**AN IMPORTANT FIX**


skabashnyuk:
ok. @tareqhs  you can merge this pr

PR #729, opened 2016-08-02
A single flow for all outgoing HTTP requests

tareksha:
phase 1 - builder, runner, all json requests

An injected interface, HttpRequestFactory, allows managing all outgoing
HTTP requests. HttpJsonRequest factory should use it internally. The
defult DefaultHttpJsonRequest as reusable as it accepts the general HTTP
factory in its constructor. Extensions can replace HttpJsonRequestFactory
and provide DefaultHttpJsonRequest as-is, given a different impl.

Change-Id: I2f334b7bb133b3f136d2b8e5b18301602f427355

### DIFF
--- a/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteTask.java
+++ b/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteTask.java
@@ -17,20 +17,16 @@ import org.eclipse.che.api.builder.internal.Constants;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.rest.HttpJsonRequestFactory;
 import org.eclipse.che.api.core.rest.HttpOutputMessage;
+import org.eclipse.che.api.core.rest.HttpResponse;
 import org.eclipse.che.api.core.rest.shared.dto.Link;
-import org.eclipse.che.commons.env.EnvironmentContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.ws.rs.HttpMethod;
 import javax.ws.rs.core.HttpHeaders;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
 import static org.eclipse.che.api.builder.BuilderUtils.builderRequest;
 
 /**
@@ -217,15 +213,7 @@ public class RemoteTask {
     }
 
     private void readFromUrl(String url, final HttpOutputMessage output) throws IOException {
-        final HttpURLConnection conn = (HttpURLConnection)new URL(url).openConnection();
-        conn.setConnectTimeout(60 * 1000);
-        conn.setReadTimeout(60 * 1000);
-        conn.setRequestMethod(HttpMethod.GET);
-        final EnvironmentContext context = EnvironmentContext.getCurrent();
-        if (context.getUser() != null && context.getUser().getToken() != null) {
-            conn.setRequestProperty(HttpHeaders.AUTHORIZATION, context.getUser().getToken());
-        }
-        try {
+        try (HttpResponse conn = requestFactory.target(url).setTimeout(60 * 1000).request()) {
             output.setStatus(conn.getResponseCode());
             final String contentType = conn.getContentType();
             if (contentType != null) {
@@ -237,13 +225,10 @@ public class RemoteTask {
                 output.addHttpHeader(HttpHeaders.CONTENT_DISPOSITION, contentDisposition);
             }
 
-            try (InputStream in = firstNonNull(conn.getErrorStream(), conn.getInputStream());
+            try (InputStream in = conn.getInputStream();
                  OutputStream out = output.getOutputStream()) {
                 ByteStreams.copy(in, out);
             }
-
-        } finally {
-            conn.disconnect();
         }
     }
 }
--- a/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/SourcesManagerImpl.java
+++ b/platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/SourcesManagerImpl.java
@@ -11,8 +11,12 @@
 package org.eclipse.che.api.builder.internal;
 
 import org.eclipse.che.api.builder.dto.BaseBuilderRequest;
+import org.eclipse.che.api.core.rest.HttpJsonRequest;
+import org.eclipse.che.api.core.rest.HttpJsonRequestFactory;
+import org.eclipse.che.api.core.rest.HttpRequest;
+import org.eclipse.che.api.core.rest.HttpRequest.BodyWriter;
+import org.eclipse.che.api.core.rest.HttpResponse;
 import org.eclipse.che.api.core.util.ValueHolder;
-import org.eclipse.che.commons.env.EnvironmentContext;
 import org.eclipse.che.commons.json.JsonHelper;
 import org.eclipse.che.commons.json.JsonParseException;
 import org.eclipse.che.commons.lang.IoUtil;
@@ -29,6 +33,7 @@ import org.everrest.core.impl.header.HeaderParameterParser;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.BufferedWriter;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -40,7 +45,9 @@ import java.io.OutputStreamWriter;
 import java.io.StringReader;
 import java.io.Writer;
 import java.net.HttpURLConnection;
-import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.text.ParseException;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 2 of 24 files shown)


tareksha:
@skabashnyuk @evoevodin please review

platform-api/che-core-api-runner/src/main/java/org/eclipse/che/api/runner/internal/Runner.java
```diff
@@ -90,6 +92,10 @@ public boolean isValid(DeploymentSources deployment) {
     private ScheduledExecutorService cleanScheduler;
     private java.io.File             deployDirectory;
 
+    @Inject
+    private DownloadPlugin theDownloadPlugin;
```
garagatyi, line 96: Can you explain why we still need old field, looks like it is unused.

platform-api/che-core-api-runner/src/main/java/org/eclipse/che/api/runner/RemoteRunnerProcess.java
```diff
@@ -128,15 +120,7 @@ public void readRecipeFile(OutputProvider output) throws IOException, RunnerExce
...
-        final HttpURLConnection conn = (HttpURLConnection)new URL(url).openConnection();
-        conn.setConnectTimeout(60 * 1000);
-        conn.setReadTimeout(60 * 1000);
-        conn.setRequestMethod(method);
-        final EnvironmentContext context = EnvironmentContext.getCurrent();
-        if (context.getUser() != null && context.getUser().getToken() != null) {
-            conn.setRequestProperty(HttpHeaders.AUTHORIZATION, context.getUser().getToken());
-        }
-        try {
+        try (HttpResponse conn = requestFactory.request(0, url, method, null, null, null)) {
```
garagatyi: I think it is better to keep timeout value to avoid any changes in app behavior

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpRequestFactory.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param body
+     *            A {@link BodyWriter} to use for writing the body. (optional)
+     * @param headers
+     *            A map of the header values. (optional)
+     * @param queryParams
+     *            A list of the query parameters to append to the URL. (optional).
+     * @return
+     * @throws IOException
+     */
+    HttpResponse request(int timeout, String url, String method, BodyWriter body, Map<String, String> headers,
```
garagatyi: For me it is very hard to read code with usage of method that has so many parameters. HttpJsonRequestFactory uses another approach with builder like API which is much more readable in several cases.
What do you think about changing general request factory to similar approach?

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteTask.java
```diff
@@ -217,15 +214,7 @@ public void downloadResultArchive(String archType, HttpOutputMessage output) thr
...
-        final HttpURLConnection conn = (HttpURLConnection)new URL(url).openConnection();
-        conn.setConnectTimeout(60 * 1000);
-        conn.setReadTimeout(60 * 1000);
-        conn.setRequestMethod(HttpMethod.GET);
-        final EnvironmentContext context = EnvironmentContext.getCurrent();
-        if (context.getUser() != null && context.getUser().getToken() != null) {
-            conn.setRequestProperty(HttpHeaders.AUTHORIZATION, context.getUser().getToken());
-        }
-        try {
+        try (HttpResponse conn = requestFactory.request(0, url, null, null, null, null)) {
```
garagatyi: Can you leave old timeout to avoid any changes in app behavior?

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/RemoteTask.java
```diff
@@ -217,15 +214,7 @@ public void downloadResultArchive(String archType, HttpOutputMessage output) thr
...
             output.setStatus(conn.getResponseCode());
```
garagatyi, line 217: I don't see where authorization headers is set. Can you point me?

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/SourcesManagerImpl.java
```diff
@@ -202,8 +205,6 @@ public void write(byte[] b) throws IOException {
     };
 
     private void download(String downloadUrl, java.io.File downloadTo, File directory) throws IOException {
-        HttpURLConnection conn = null;
-        try {
             final LinkedList<java.io.File> q = new LinkedList<>();
```
garagatyi: Since try is removed looks like indentation is too big

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
```diff
@@ -181,48 +182,24 @@ protected DefaultHttpJsonResponse doRequest(int timeout,
                                                                                UnauthorizedException,
                                                                                ConflictException,
                                                                                BadRequestException {
-        final String authToken = getAuthenticationToken();
-        final boolean hasQueryParams = parameters != null && !parameters.isEmpty();
-        if (hasQueryParams || authToken != null) {
```
garagatyi, line 186: Since authentication is dropped this change will change app behavior. Looks like it is not very safe for stability of the project

garagatyi, line 186: Sorry, now I see that you've moved this code.

platform-api/che-core-api-runner/src/main/java/org/eclipse/che/api/runner/internal/Runner.java
tareksha, line 96: @garagatyi the old field is final protected, not private. i saw it too risky to just remove it or hide non-incrementally.

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpRequestFactory.java
tareksha: @garagatyi correct, but DefaultHttpJsonRequest uses a non-streaming approach and consumes the body of both requests and responses as full strings, which is not suitable for general HTTP streaming. the change is intended to minimize changes by exposing a common denominator. otherwise the change will be much bigger.

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/SourcesManagerImpl.java
tareksha: @garagatyi I did this to avoid bumping the whole file to the left and affect only the specific lines that changed. I can un-indent it.

platform-api/che-core-api-runner/src/main/java/org/eclipse/che/api/runner/internal/Runner.java
garagatyi, line 96: I didn't notice that, thanks.

platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/HttpRequestFactory.java
garagatyi: Is it possible to use builder style API of http client with body writer/reader that will implement streaming approach?

platform-api/che-core-api-builder/src/main/java/org/eclipse/che/api/builder/internal/SourcesManagerImpl.java
garagatyi: I think github will show that changes are in indentation only

tareksha:
@garagatyi  the change now uses the same builder pattern. please review

tareqhs:
@garagatyi @skabashnyuk @vparfonov @evoevodin it passes all tests

skabashnyuk:
@tareqhs Quick response is -1 to make HttpJsonRequest something not related to json. We understand that you want to make some tooling to make general request but making it in HttpJsonRequest is not that we want to achieve. I hope @garagatyi @evoevodin  can add more to this statement.

voievodin:
Hi @tareqhs.
Generally speaking `HttpJsonRequest` was designed to work with json only, there are several reasons for that
- 90% requests are requests between our API services where json is a main format, so it fits almost all the needs
- when the tool is json specialized we can add useful methods related to DTO processing, which is also json based, so the tool becomes even more convenient and usable.

The idea of you changes is clear for me and makes total sense, but the implementation strategy should be different, as `HttpJsonRequest` becomes not really Json and the client of the class may be confused as the class contains general and json request logic mixed.

As for me the best solution here is to create  general class like `HttpRequest` with all the general methods for requesting stuff through http and then change the current `HttpJsonRequest` implementation to use `HttpRequest` in this way the logic is clearly separated between components.

tareksha:
hi @evoevodin @skabashnyuk ,

Well, that was the original approach in the PR but I tried to minimize code changes. The main issue is not whether to extend `HttpJsonRequest` itself or not, but rather the factory `HttpJsonRequestFactory`. I'm seeking a solution that doesn't require injecting a new factory everywhere yet again, which conceptually is a superset of `HttpJsonRequestFactory`.

This is my suggestion : `HttpJsonRequestFactory` extends a general `HttpRequestFactory` that has general request methods, so that the existing injections stay identical except for receiving the additional HTTP APIs that are in turn used instead of `HttpURLConnection` where needed.

Agree?

PR #730, opened 2016-08-03
Install maven 3.x before ci starts

tareksha:


skabashnyuk:
ok. Squash commits before merge please.

tareksha:
@skabashnyuk  had to disable . travis provides maven 3.2 and Che requires 3.3, which is failing to install and run correctly in travis . will enable once fixed, meanwhile tests should run locally

skabashnyuk:
@tareqhs  What we can do?

tareksha:
@skabashnyuk we'll have to investigate into this separately without affecting 3.x i guess, otherwise we will get constant false negatives in all pull requests. if you have any immediate solutions this can be very useful.

PR #732, opened 2016-08-08
Added contribution templates and howtos

skabashnyuk:
Used templates from che.


skabashnyuk:
@tareqhs  wdyt?

tareksha:
@skabashnyuk looks good for me

PR #735, opened 2016-08-10
Vfs improve

tareksha:
### What does this PR do?
- Use permissions enum directly
- Rewrite hasPermission() to be easier to read and maintain
- Delay VirtualFileSystemUser retrieval to avoid in case unnecessary
- AutoCloseable and resource try to safer lock releasing in FSMountPoint

Maintainability and extensibility of FSMountPoint
### Tests written?

No. Behavior kept as-is and tests pass.

### DIFF
--- a/che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
+++ b/che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
@@ -21,6 +21,7 @@ import org.eclipse.che.api.vfs.server.LazyIterator;
 import org.eclipse.che.api.vfs.server.MountPoint;
 import org.eclipse.che.api.vfs.server.Path;
 import org.eclipse.che.api.vfs.server.PathLockFactory;
+import org.eclipse.che.api.vfs.server.PathLockFactory.PathLock;
 import org.eclipse.che.api.vfs.server.SystemPathsFilter;
 import org.eclipse.che.api.vfs.server.VirtualFile;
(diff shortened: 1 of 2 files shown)


tareksha:
@skabashnyuk @benoitf @vparfonov please review

che-core-vfs-impl/src/main/java/org/eclipse/che/vfs/impl/fs/FSMountPoint.java
```diff
@@ -1445,43 +1417,54 @@ void updateACL(VirtualFileImpl virtualFile, List<AccessControlEntry> acl, boolea
...
-                        }
-                    }
-                }
-                final Principal anyPrincipal = DtoFactory.getInstance().createDto(Principal.class)
-                                                         .withName(VirtualFileSystemInfo.ANY_PRINCIPAL)
-                                                         .withType(Principal.Type.USER);
-                userPermissions = accessControlList.getPermissions(anyPrincipal);
-                return userPermissions != null && (userPermissions.contains(p) || userPermissions.contains(BasicPermissions.ALL.value()));
+        AccessControlList accessControlList;
+        while (true) {
```
benoitf, line 1424: could we avoid such pattern of 'while true' ?

tareksha, line 1424: @benoitf the loop body has multiple terminal points which is somewhat restrictive because i intend to reduce double checks to zero. That's why I wrote the loop as simple as possible with verbose documentation

skabashnyuk:
ok

PR #706, opened 2016-04-24
fix bug with folders cleanup not being performed correctly

freedomp:
fix bug with folders cleanup not being performed correctly
on project import and upload zip project api

Signed-off-by: Dror Cohen dror.cohen@sap.com

PR #737, opened 2016-09-12
Fix of Import failure - Cleanup folder created during Import

MeiravDaniel:
### What does this PR do?

This pull request replaces this pull request - 
https://github.com/codenvy/che-core/pull/706
It add a check if the root folder of the import exist before import
If it does and the import failed - it deletes the folder

The problem of having a "leftover" folder when import failed and there was no folder before
### Previous Behavior

The folder was created and remain even if the import failed
### New Behavior

the folder is deleted on failure
### Tests written?

No
### Docs requirements?

No

Please review [Che's Contributing Guide](https://github.com/eclipse/che/CONTRIBUTING.MD) for best practices.

https://github.com/codenvy/che-core/pull/706/files
I would like to mention that the comment on "newDto" and Collection Map
was not fixed because the origin change did not change the body of the method
It only wrap them by try {} catch , so if a folder was created during the
Import - It will be deleted on failure

Signed-off-by: i028422 meirav.daniel@sap.com

### DIFF
--- a/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
+++ b/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
@@ -806,15 +806,25 @@ public class ProjectService extends Service {
         // Not all importers uses virtual file system API. In this case virtual file system API doesn't get events and isn't able to set
         // correct creation time. Need do it manually.
         long creationDate = System.currentTimeMillis();
+        boolean folderExistedBeforeCommand = virtualEntryExists(workspace, path);
         VirtualFileEntry virtualFile = getVirtualFile(workspace, path, force);
 
-        final FolderEntry baseProjectFolder = (FolderEntry)virtualFile;
-        importer.importSources(baseProjectFolder, projectSource.getLocation(), projectSource.getParameters(), outputOutputConsumerFactory);
-
-        //project source already imported going to configure project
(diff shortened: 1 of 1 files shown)


MeiravDaniel:
@tareqhs could you please review?
Thanks,
Meirav

tareksha (commented):
Does it pass all unit tests?

MeiravDaniel:
Yes.It does

platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -806,15 +806,25 @@ public ImportResponse importProject(@ApiParam(value = "Workspace ID", required =
...
         VirtualFileEntry virtualFile = getVirtualFile(workspace, path, force);
 
-        final FolderEntry baseProjectFolder = (FolderEntry)virtualFile;
-        importer.importSources(baseProjectFolder, projectSource.getLocation(), projectSource.getParameters(), outputOutputConsumerFactory);
-
-        //project source already imported going to configure project
-        return configureProject(importProject, baseProjectFolder, workspace, creationDate);
+        FolderEntry baseProjectFolder = null;
+        try {
+        	 baseProjectFolder = (FolderEntry)virtualFile;
```
skabashnyuk, line 814: wrong formatting

skabashnyuk (approved):

skabashnyuk:
@tareqhs  can you confirm that formation is fixed?

PR #739, opened 2016-11-09
Expose optional java.io.File from VirtualFile

tareksha:

Alternative VFS providers could not integrate correctly into existing Che services, especially the Git API.

### Previous Behavior
Several areas in the code failred during runtime due to direct cast to ``VirtualFileImpl``, although the given ``VirtualFile`` instance has a different implementation provided from an alternative VFS implementation.

### New Behavior
``VirtualFile`` exposes a new method ``java.io.File getIoFile()`` that returns a concrete ``File`` if such file is available, otherwise null. Che services use this method instead of casting to ``VirtualFileImpl`` unsafely, failing with proper exceptions if a ``File`` is unexpectedly not available.

avoid direct casts to VirtualFileImpl and expose the required function
instead, allowing alternative VFS provider to integrate flawlessly in the
Che platform with any desired java.io.File delegations

Change-Id: I6a6e7ef65279c6aa2714dcedb75dcb4dce5e50ff


tareksha:
@skabashnyuk please review

skabashnyuk (approved):

